Issue #30, opened 2016-04-19
[F] carry.out includes dplyr::select style renaming


dpastoor:
example scenario, I am passing in a dataset with DV, and want to simulate against it, but also retain the original DV column to carry through. As of now, it will do so, however will result in 2 columns named DV 

![image](https://cloud.githubusercontent.com/assets/3196313/14645701/b613dc56-0624-11e6-9364-8fa661128d50.png)

```
model %>% 
idata_set(param_data) %>%
  data_set(my_data ) %>% 
  carry.out(evid, WT, RDV = DV) %>%
  mrgsim(end= max(my_data$time) + 12,delta=0.1,digits=4)
```

such that the column DV would be carried forward and renamed RDV

kylebmetrum:
I think we'll be able to rename items in `Request()` by the same mechanism as well (change the name of a tabled item right after simulating and before returning from `mrgsim`).

kylebmetrum:
Not committed yet.


``` r
see(ind_model)
```

    ## 
    ## Model file:  ind_model.cpp 
    ## 
    ## $PARAM CL= 1.2, VC=14.3
    ## $CMT GUT CENT
    ## $ADVAN2
    ## $MAIN
    ## pred_CL = CL;
    ## pred_V = VC;
    ## $TABLE table(DV) = CENT/pred_V; 
    ## 

``` r
out <- 
  ind_model %>%
  data_set(ex_id_data) %>% 
  carry.out(BAR = FOO, CMT = cmt, EVENT_ID = evid, addl) %>% 
  Req(CENTRAL = CENT, DV) %>%
  mrgsim(delta=0.1, digits=4) 

out
```

    ## Model:  ind_model.cpp 
    ## Dim:    244 x 8 
    ## Time:   0 to 24 
    ## ID:     1 
    ##      ID time CMT EVENT_ID addl BAR CENTRAL      DV
    ## [1,]  1  0.0   0        0    0   1  0.0000 0.00000
    ## [2,]  1  0.0   1        2    0   1  0.0000 0.00000
    ## [3,]  1  0.1   0        0    0   1  0.9958 0.06964
    ## [4,]  1  0.2   0        0    0   1  1.9830 0.13870
    ## [5,]  1  0.3   0        0    0   1  2.9630 0.20720
    ## [6,]  1  0.4   0        0    0   1  3.9340 0.27510
    ## [7,]  1  0.5   0        0    0   1  4.8970 0.34240
[...]

PR #32, opened 2016-04-21
create rename_cols function with tests

dpastoor:
Refers to discussion at #30 could be used to rename columns by using rename columns, 
thus the proposed method could have a rename function added as below

```
setMethod("data_set",c("mrgmod", "data.frame"), function(x,data,subset=TRUE,select=TRUE, rename=NULL,...) {
    if(exists("data", x@args)) stop("data already has been set.")
    if(!missing(subset)) data <- dplyr::filter_(data,.dots=lazy(subset))
    if(!missing(select)) data <- dplyr::select_(data,.dots=lazy(select))
    data <- dplyr::rename_(data,.dots=lazy_dots(...))
    if(nrow(data) ==0) stop("Zero rows in data after filtering.", call.=FALSE)
   if(!is.null(rename)) data <- rename_cols(data, rename)
    data <- mrgindata(m=x,x=data,...)
    x@args <- merge(x@args,list(data=data), strict=FALSE)
    x
})
```


dpastoor:
note - I chose to keep it unexported as if they want to do renaming "themselves" with NSE or some other way, it'd probably just be best to actually use the dplyr verbs directly piping them before it hits data_set like `... data_set(my_data %>% rename(somecol=oldcol))...` giving the `rename` arg the case of batch renaming, especially in a programmatic context where you just want to pass in a string vector

dpastoor:
well this PR won't likely merge - I am getting this error 
image

I can dig around your build process, it is much more complex than I am used to, I usually just devtools::document() --> devtools::test()

That said, the util function tests do pass

image

it just seems that they aren't being fully integrated into the package

dpastoor:
ok lets see if exposing the rename_cols via `:::` will do the trick

dpastoor:
In light of discussion on #30 feel free to close/reject this, I'll keep it open for now just in case you want to reference it for any other internals

dpastoor:
@kylebmetrum ah blarg it just updated this PR - do you have any interest in keeping the rename cols function? else I can rebase, eliminate the code related to this PR and resubmit with the refactoring of  `check_pred_symbols`

PR #36, opened 2016-04-29
fix rename_cols export issue in tests and catch extra documentation

dpastoor:


dpastoor:
@kylebmetrum 🎆 victory - we are in sync

PR #108, opened 2016-08-21
Annotated blocks; save parsed objects in environment

kylebmetrum:
About this PR:

# New Features
- Added the ability to "annotate" selected code blocks.  Annotation includes a description, unit specification, and "option" list for the following code blocks: `$PARAM`, `$FIXED`, `$THETA`, `$CMT`, `$VCMT`, and `$INIT`.  Use the regular block names, but add `annotated=TRUE` block option.  

# Bugs Fixed
- Bug fixed when `ALAG_CMT` was very small (smaller than `.Machine$double.eps`) such that the lag time was zero.  This causes a problem in record sorting.  After the fix, the lag time is only implemented if `ALAG_CMT` is greater than `mindt` (see `?mrgsim`). (#109)

# Under the hood
- Several changes in how code blocks are parsed.  
  - Each code block is assigned a `pos` attribute, indicating the numerical order of that code block in the model specification
  - We are now passing an `R` environment into block handler functions.  For certain handler functions, when `R` objects are created in the handlers, they are stored in lists inside the parse environment using `pos` (see above). After all code blocks are parsed, the `R` objects are retrieved from the environment.


kylebmetrum:
@vjd @dpastoor 

Lines starting with `>>` indicate that block options are being specified.  So it could be:

```r
$PARAM >> annotated=TRUE
...
```

or

```r
$PARAM
>> annotated=TRUE
...
```


A model specification file could look like this (both annotated and non-annotated blocks shown).

```r
$PARAM  >> annotated=TRUE
CL: 1 : Yes (L/hr)
VC: 2 : NO (L)

$PARAM A=1, B=2

$CMT >> annotated=TRUE
CENT   : Central (ng)
PERIPH : Peripheral compartment (ng)

$CMT D E F

$FIXED >> annotated=TRUE
g    : 9.8 : From physics class (m/s^2) 
conv : 1.2 : Some conversion factor (.)

$FIXED d = 2
```

For now, I just collect information for blocks that are annotated and make it available in some close-to-useable form.  Down the road, we can get more sophisticated about reconciling what is annotated and what is not.  It should be easy to specify an external file that gets parsed on model compile to provide the same information that you would get in an `annotated` block.


```r
> mod <- mcode("hey", code)
> 
> mod@annot
[[1]]
  block name descr unit options
1 PARAM   CL   Yes L/hr       .
2 PARAM   VC    NO    L       .

[[2]]
  block   name                  descr unit options
1   CMT   CENT                Central   ng       .
2   CMT PERIPH Peripheral compartment   ng       .

[[3]]
  block name                  descr  unit options
1 FIXED    g     From physics class m/s^2       .
2 FIXED conv Some conversion factor     .       .

```

kylebmetrum:
Latest commit: for (potentially) annotated blocks, the options are scraped first and passed as arguments to secondary handler functions.  So options can be partially matched:


```r
$PARAM >> annot=TRUE
CL: 12 : Clearance (L/hr)
VC: 40 : Volume (L)
```
etc ...

kylebmetrum:
More examples in `test-annot`
https://github.com/metrumresearchgroup/mrgsolve/blob/08914cfb34aed54250c18a697bea6bdb00c06bd4/rdev/tests/testthat/test-annot.R

Issue #127, opened 2016-09-20
change annotation syntax to be more 'macro-like'


dpastoor:
in typing out #121 discussions, I realized, given we (potentially) have a number of settings that are either TRUE/FALSE, those boolean values could be handled in a more macro/decorator-like syntax

instead of

```
$OMEGA 
annotated = TRUE 
...
```

any booleans could be proceeded by `@`

```
$OMEGA 
@annotated 
...
```

```
$OMEGA
@annotated 
@block
...
```

I would presume it would also actually be easier to parse too, since you don't have to handle full expressions just look for what `@` macros were present

kylebmetrum:
This was in cd4e9e6.  In the commit, `@` macros are only found in designated option lines (starting with `>>`).  We could look elsewhere in the block, but I'd like to migrate away from bare options in a block.  It's not that we couldn't parse it out ... just seems like more work / code to have to check anywhere in a block for the options.

```r
$OMEGA >> @annotated, @block
ECL: 1 : ETA on CL
EVC: 0.001 1: Eta on VC

$OMEGA >> @annotated, @block
EKIN:  ETA on KIN
EEC50: ETA on EC50

1  0.001 3
```

dpastoor:
ah - i was thinking just make them more explicit in there location to reduce headaches - namely they just be right below the $<block> statement

kylebmetrum:
Any objection to insisting on the `>>` tag for options?  

We're already picking that out and I feel like we can make this even more robust if we can first look for `>>` and then dice that up.  

Right now you can still do this
```r
$OMEGA
block = TRUE
1 2 3 4
```

This goes back a long time ago.  Basically we grep out `=` and take the whole line as comma-separated options.  We can't do that for `$PARAM` or other blocks where the data is `name=value`.  So I came up with `>>`.   I didn't really like `>>` at first but it's grown on me and seems more natural now when you put options right after the block name.  I'd like to get to the point where we phase out any option that is not preceded by `>>`.  I feel like we'll be able to do more with those options.  In the end, it will be less to look for.

dpastoor:
I don't have a problem with it, the 'confusing' thing is `>>` a per line thing

```
$OMEGA
>> @annotate
>> @block
```

or just at the beginning of annotations. Also, does it need to be on the same line as `$statement` or can it be separated?


For example, my personal styling is 1 per line, this makes it easy to diff things by line

```
$OMEGA
>> @annotate
>> @block
```

to 

```
[...]

PR #129, opened 2016-09-27
rm table(); add qsim; add -at- macros for annotating models

kylebmetrum:


kylebmetrum:
This turned out to be a bunch of changes after deprecating `table()` that I held until that was ready to go. 


1. `table(name) = value;` is now deprecated.  An error message will be generated.  To get derived values into the simulated output, users should save the value to `double` and list that name in `$CAPTURE`.  
1. Added `@` macros for adding options to code blocks as suggested and developed in #127.  
1. When `matlist` labels are not the appropriate length, `mrgsolve` includes the actual and expected length in the error message. 
1. User variables in the model are now stored in an unnamed namespace so that they are local to the file and won't be able to be found in the shared object.
1.  Added `qsim` function that is experimental lightweight, no-frills simulation function.  This is not ready for production and needs documentation, but it works.

PR #131, opened 2016-10-05
mrgsolve:::render development

kylebmetrum:


rdev/R/utils.R
```diff
@@ -196,41 +196,60 @@ expand.ev <- function(...) {
 is.numeric.data.frame <- function(x) sapply(x, is.numeric)
 
 
+
+##' Render an \code{mrgsolve} model.
+##' 
+##' @param x model object
+##' @param file file name to use for building; should have \code{.Rmd} extension
+##' @param quiet not used
+##' @param ... not used
```
dpastoor: you could pass ... to the rmd render function to add some additional customization

rdev/R/utils.R
```diff
@@ -196,41 +196,60 @@ expand.ev <- function(...) {
...
   
   cat(file=file, txt, sep="\n")
   cat(file=file, 
-      "```{r,echo=FALSE,comment=''}", 
-      paste0("x <- readRDS(file=\"",anot, "\")"),
-      "```",
       "# Model details",
-      "```{r,echo=FALSE,comment=''}", 
-      "print(x$details)",
+      "```{r,eval=FALSE,echo=TRUE,comment=''}",
```
dpastoor: given everything is eval=F none of this code will run? Maybe I'm missing the objective of the render function?

rdev/R/utils.R
```diff
@@ -196,41 +196,60 @@ expand.ev <- function(...) {
...
+
+##' Render an \code{mrgsolve} model.
+##' 
+##' @param x model object
+##' @param file file name to use for building; should have \code{.Rmd} extension
+##' @param quiet not used
+##' @param ... not used
+##' 
 render <- function(x,file=tempfile(fileext=".Rmd"),quiet=TRUE,...) {
   stopifnot(requireNamespace("rmarkdown"))
```
dpastoor: you might consider a more informative error message

```
 if(!requireNamespace("rmarkdown", quietly = T)) {
	    stop("need rmarkdown to use this function, please install via install.packages("rmarkdown") ")
}
```

rdev/R/utils.R
```diff
@@ -196,41 +196,60 @@ expand.ev <- function(...) {
...
   
   for(i in seq_along(bl)) {
+    bl[[i]] <- bl[[i]][bl[[i]] !=""]
     bl[[i]] <- paste(bl[[i]], collapse="\n") 
   }
-  saveRDS(file=anot, list(details=details(x),code=x@code,bl=bl))
+
+
+  w <- options()$width
+  options(width=120)
```
dpastoor: is this code to not truncate the printed table when capturing the output or something?

rdev/R/utils.R
```diff
@@ -252,9 +266,14 @@ render <- function(x,file=tempfile(fileext=".Rmd"),quiet=TRUE,...) {
       capture.output(cat(bl$TABLE,sep="\n")),
       "```",
       sep="\n",append=TRUE)
-  ret <- rmarkdown::render(file,quiet=quiet,output_format="pdf_document")
-  file.copy(ret,file.path(getwd(),basename(ret)))
-  return(ret)
+  
+  if(build) {
+    ret <- rmarkdown::render(file,quiet=quiet,output_format="pdf_document")
```
dpastoor: would this be better to wrap in a tryCatch block, so you can bubble up times when the rendering fails (for whatever reason it might)?

rdev/R/utils.R
kylebmetrum: Yeah ... `R` defaults to 80 (but I think `Rstudio` might change it to 120)?  Anyway, output gets wrapped way too early.  I might need to fiddle with this.

rdev/R/utils.R
kylebmetrum: Good idea.

rdev/R/utils.R
kylebmetrum: yes ... probably should be.  I took control of the `render` options to simplify; but seems a bit restrictive.

rdev/R/utils.R
kylebmetrum: Maybe don't need that collapse business on the code blocks any more ... from a previous iteration.

PR #142, opened 2016-10-14
Adding mread_cache and mcode_cache 

kylebmetrum:
These are wrappers for `mread` and `mcode` that cache the model object, allowing users to re-instantiate the model later on without compiling.  

This needs to be tested on `windows` before merging; I suspect that the shared object needs to be explicitly unloaded before trying to load the cached model.


dpastoor:
I can test early next week on Windows,  prob Monday. Do you have any tests in or just useability

PR #143, opened 2016-10-14
add mread_cache and mcode_cache

kylebmetrum:
- Also annotated some modlib models 

These are wrappers for `mread` and `mcode` that cache the model object, allowing users to re-instantiate the model later on without compiling.  

This needs to be tested on `windows` before merging; I suspect that the shared object needs to be explicitly unloaded before trying to load the cached model.


kylebmetrum:
@dpastoor Sure ... if you want to dig into it.  No tests yet;  I will add.  But if you want to look at the logic of the cache: basically, the cache file has to exist, the shared-object (uniquely-named) from the previous (successful) compile needs to still exist, and the MD5sum from the model specification file needs to be the same as in the cached object otherwise the model is rebuilt.  Maybe it's overkill ... but I can think of a bunch of ways this could get mucked up;  I'd rather be safer. 

I'm 99.99% sure this will crump on Windows;  Need a line like this: https://github.com/metrumresearchgroup/mrgsolve/blob/master/rdev/R/complog.R#L52
before reloading the shared object.  Right now, that is wrapped in a larger function that does too much (like deleting shared object files ... which will ruin the cache).  

Anyway ... I don't think there is a desperate need for this and it's not *that* hard to roll your own cache.  But lots of people ask about it and I can see where it would come in handy when you're running a shiny app ... for me, its convenience of not having to arrange permission to write somewhere when loading the app.  If feel like these wrapper functions keep it out of the way from the usual, intended work flow of just compiling via `mread` without asking users who do want to cache to do too much.

kylebmetrum:
I already had this covered in `loadso`.  It should work on Windows.

https://github.com/metrumresearchgroup/mrgsolve/blob/master/rdev/R/methods.R#L23-L31

kylebmetrum:
But I think there may still be an issue with unloading and re-loading a shared object of the same name.  

I wonder if it would be safer to:
  1.  Unload the shared objec 
  1.  re-name the shared object (file)
  1.  Load the shared object (under new name)  
  1. Update and re-cache the model object

It seems like a lot, but reading over `dyn.load` and `dyn.unload` help files makes me think it should probably happen that way.  Or ... only try to load the shared object if nothing is loaded.  That's probably simpler.

dpastoor:
Just for documentation's sake, as confirmed with kyle offline, I also tested this on a Windows 10 Pro desktop with R 3.3.1 and had no problem reading from the cache after the first load, as expected

PR #166, opened 2016-11-24
add an as_data_frame method to convert to tibble

dpastoor:
As tibble is what is directly driving the as.tbl and as_data_frame methods of dplyr. The reason that it goes through as.data.frame is as_data_frame sometimes runs into issues with direct coercion, especially from matrices.


dpastoor:
Also, as I was already using the replicate branch when I made this change - it was going to be a mess to try to merge this into master, so I figured this could just be incorporated when the replicate branch gets merged

kylebmetrum:
Just so I can understand: does this get us something different than this 

https://github.com/metrumresearchgroup/mrgsolve/blob/8e7e273a66c7ed24c163cc0e37e498cff3f4fe0c/rdev/R/mrgsims.R#L128-L130

kylebmetrum:
The `as.tbl.mrgsims` stuff was added before `tibble` was it's own package.  Maybe `dplyr::as.tbl` isn't the right thing to do anymore?

dpastoor:
correct - tibble is the more robust implementation (it was the dplyr implementation extracted out with now more development) and dplyr is moving to just alias the tibble stuff directly.

Regardless, both should give you the same output, its just different aliases.

quote from hadley basically the same question as you

> No – it’s just that the dplyr stuff is moving into a separate package. (And tibble includes a whole lot of bug fixes and small stuff that aren’t in current dplyr)

PR #175, opened 2017-01-07
Changes in 2017

kylebmetrum:
As usual, a variety of stuff.


vjd:
Just went through the details and realized that there are so many incredible features since I last worked with mrgsolve about 4 months ago. This is a great beginning to 2017!

kylebmetrum:
I tweeted out a bunch of stuff today and making serious effort to get this documented. @vjd @dpastoor

vjd:
I am actually testing the $ENV since this am.. will keep you posted!

mrgsolve:
Great!  Feedback please ... that `covset` stuff has been about 10 different things.  I'm finally feeling ok about how it looks and the implementation too. I'm not hugely invested in doing all of this "inside" the model, so tried to allow user to decide what gets embedded into the model and what doesn't.  Maybe a little more work that way, but hopefully the flexibility pays off.  And just settled on the "form" of the thing ... probably lots to do in terms of error checking etc. It would also be nice to add some expressions or formulae that just `mutate` columns in the data set.

PR #176, opened 2017-01-07
provide documentation website for mrgsolve functions

dpastoor:


dpastoor:
docs/ folder provides a site that looks like:

image


with reference showing the functions

image

and clicking on a function gives documentation

image


It is built using https://github.com/hadley/pkgdown

this would allow you to set github to serve from the docs/ folder so this could be hosted through github pages

dpastoor:
of note - 

1) I had to switch to Authors@R syntax to build - while at it I added myself as a contributor, hope that is ok

2) The readme images to not link properly ATM. If this is something you'd want to use going forward we'd need to adjust the knitr render settings so it will show up properly in both the readme on github, and in the docs/ folder

3) I did this from the tip of the `dev` branch, so you could merge this more easily.

kylebmetrum:
http://mrgsolve.github.io/docs/ @dpastoor

kylebmetrum:
I'm going to put a pkgdown target in the Makefile ... so make pkgdown will build the site.  I'll have it copy the README images to the proper location to display on the website.  But for some reason, `pkgdown::build_site` isn't working for me.

```r
Error in basename(data$path) : a character vector argument expected
```

kylebmetrum:
This all is really awesome.

kylebmetrum:
I like this, too, because I can easily browse the docs and see where the holes are (there are many).

dpastoor:
ah yeah - about that error - I wanted to just push it up to see if this was a reasonable track for you before dug to much. The bug comes from _something_ weird happening when pkgdown attempts to render the README.Rmd

I ran each of the sub-function separately (thus also needing to add the Authors@R), and was also stuck there for a while. Tracked it down to the `pkgdown::build_home()` step. 

So - my very temp solution was to manually knit the Readme.Rmd to Readme.md, delete the readme.Rmd so `build_home()` instead just grabbed the already rendered `.md` then revert the deletion by reverting the deletion via git..... painful to say the least.

I then further tracked down the issue to the `mcode()` chunk as being the offending chunk, however decided there were also some issues with even understanding how pkgdown would copy (or not) the proper image file links, before I dug further.

So - no direct solution yet - but seemed like a prime thing to bounce off hadley et al. at Rstudio conf next week.

kylebmetrum:
This is why the render wasn't working:
https://github.com/kylebmetrum/pkgdown/commit/336e73efeb7b5c8f15fa77a5f822d6f184ada337

I create an object called `data` in the `Rmd` and it clobbers the `data` object in `build_home`.  The document gets rendered in `parent.frame()` environment.  I seriously didn't think `rmarkdown::render` worked like this ... but it sure seems like passing `envir = new.env()`  to `render` (or changing `data` to `.data` in the `README`) will fix.

kylebmetrum:
http://metrumresearchgroup.github.io/mrgsolve/

PR #189, opened 2017-01-23
disable dmutate everything for now

kylebmetrum:
Also fixes the issue if importing from tibble::as_data_frame after importing from dplyr::as_data_frame


kylebmetrum:
Taking internal support for `covset/dmutate` out of the package for now.   I want to re-think the implementation and the decision of whether to internalize `dmutate` stuff versus external package.  I want `dmutate` to be external package and not ready to accept the implementation (so far) for working it in as a Suggested package.

Issue #218, opened 2017-04-18
dplyr 0.7.1


kylebmetrum:
Hi,

This is an automated email to let you know about the release of dplyr,
which will be submitted to CRAN in the near future (May 11).

* Major changes: <https://blog.rstudio.org/2017/04/13/dplyr-0-6-0-coming-soon/>
* All changes: <https://github.com/tidyverse/dplyr/releases/tag/v0.6.0-rc>
* Backward compatibility guide:
<http://dplyr.tidyverse.org/articles/compatibility.html>

To check for potential problems, I ran `R CMD check` on your package
mrgsolve (0.8.6).

I found: 0 errors | 1 warning  | 1 note .

    checking examples ... WARNING
    Found the following significant warnings:

      Warning: 'mutate_each_' is deprecated.
      Warning: 'mutate_each_' is deprecated.
    Deprecated functions may be defunct as soon as of the next release of
    R.
    See ?Deprecated.

    checking installed package size ... NOTE
      installed size is  6.2Mb
      sub-directories of 1Mb or more:
        libs   5.0Mb

If I got an ERROR because I couldn't install your package (or one of
its dependencies), my apologies. You'll have to run the checks
yourself (unfortunately I don't have the time to diagnose installation
failures as I have to run checks on hundreds of packages).

Otherwise, please carefully look at the results. If you think I've
introduced a bug in dplyr, please file a reprex at
<https://github.com/tidyverse/dplyr/issues>. Otherwise, you'll need to
prepare an update to your package following the advice in
<http://dplyr.tidyverse.org/articles/compatibility.html>.

To get the development version of dplyr so you can run the checks
yourself, you can run:

    # install.packages("devtools")
    devtools::install_github("tidyverse/dplyr")

If you have any questions about this email, please feel free to
respond directly.

Regards,

Hadley

dpastoor:
I think my PR fixes the reason it couldn't be installed.

Regarding the `mutate_each_` --> `mutate_all`

However this will also involve the transition to the new NSE handling, so it may be worth waiting while the bugs get ironed out and R-lang gets onto CRAN so it doesn't introduce extra headache for other people.

kylebmetrum:
It looks like `dplyr 0.7.0` is on CRAN.  I installed and was able to build `mrgsolve` fine.  I'm not sure how much of the "new" stuff is in that release.  

[...]

PR #229, opened 2017-04-29
Inven PR to track changes

dpastoor:
Dialog in the PR probably more relevant than issue


dpastoor:
Sweet! Already looks great.

I wonder for the need syntax if we could consider leveraging dplyr select-style implementation

```
data_set(..., need=vars(CL:V2))
data_set(..., need=vars(everything()))

idata_set(..., need=vars(starts_with("E"))) # if every eta prefaced with E maybe
```

kylebmetrum:
That's a good idea to leverage dplyr select stuff.  That sort of thing could really make or break this.  If specifying `need` is clumsy or inflexible, this won't work well.

dpastoor:
so I'm digging around, and at least in the new version of dplyr, they are exposing a lot of the select machinery

https://github.com/tidyverse/dplyr/blob/85ad81813da7eef90b89d7d1b76c3edb807ff217/R/select-vars.R

``` r
library(dplyr)


names(Theoph)
#> [1] "Subject" "Wt"      "Dose"    "Time"    "conc"

select_vars(names(Theoph), Dose:conc)
#>   Dose   Time   conc 
#> "Dose" "Time" "conc"

select_vars(names(Theoph), -conc, -Dose)
#>   Subject        Wt      Time 
#> "Subject"      "Wt"    "Time"
```

dpastoor:
so....

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 


## we'd pull this internally from the mrgmod
## so hard coding to fake that
.data <- names(Theoph)

need <- function(...){
  select_vars(.data, ...)
}
  
need(everything())
#>   Subject        Wt      Dose      Time      conc 
#> "Subject"      "Wt"    "Dose"    "Time"    "conc"

need(Wt:conc)
#>     Wt   Dose   Time   conc 
#>   "Wt" "Dose" "Time" "conc"
```

kylebmetrum:
`select_vars_` is exposed in `0.5.0`.  This works fine for the `inventory` function where we can just take the dots and required parameters for that object.  `data_set` etc are more complicated functions.  So would it be:

```
mod %>% data_set(data, need=vars(CL:V2,KM,KA)) %>% 
```
or whatever?

kylebmetrum:
This is sweet.

```r
> code <- '$PARAM CL = 1 , V2 = 3, KA = 4, VM = 5'
> 
> mod <- mcode_cache("tst", code)
Loading model from cache.
> 
> idat <- data_frame(CL=1,V2=3,KA=4,VM=10)
> 
> inventory(mod,idat, everything())
Found all required parameters
```

kylebmetrum:
```r
> code <- '$PARAM CL = 1 , V2 = 3, KA = 4, VM = 5'
> inventory(mod,idat,CL:KA)
object is missing these parameters
 - V2
```

dpastoor:
exactly - do note though that select_vars_ (and every `_` function) is deprecated in 0.6.0.  I'm trying to think if there is a (reasonably) elegant way of supporting both 0.5 and 0.6 without knowingly using a soon-to-be-deprecated function.

Regardless - this feels cleaner - really liking it!

kylebmetrum:
I see.  Thanks for clarifying that ... and for suggesting the select stuff.  I like this a lot.  Definitely going to keep heading down this path.

kylebmetrum:
This is on the same PR unfortunately.  Based on @dpastoor suggestion in #218, changing `mutate_each` to `mutate_at` only if using the new `dplyr.  There is also conditional import (`mutate_each` or `mutate_at`).

@dpastoor should we support both `dplyr` versions and then just require `dplyr >= 0.6.0` at some point?

I haven't really gotten into `0.6.0` or anything ... there may be other gremlins in there.  But as of now, we're passing tests and check.

kylebmetrum:
Verifying that everything still works after reverting to `dplyr 0.5.0`.

dpastoor:
@kylebmetrum I am not sure regarding dplyr 0.5 + 0.6 support atm. I can definitely address this in the next couple weeks as I can dig into rlang vs lazyeval. The main thing that I am not sure of is given the new syntax with the `!!` to indicate that it is a reference to a variable vs a named value in the data, how will that support work for packages that don't understand rlang-based syntax. 

My thoughts are that it should be possible to support both, especially in the short term. The worst case scenario is once rlang is on CRAN that can be the standardized external api that mrgsolve uses, which would just directly pass through to dplyr 0.6, but internally could wrap the dplyr 0.5 syntax as well.

Let me check back in around the 15-20th and I can give some more conclusive thoughts. This is something that I am generally wondering about beyond mrgsolve so it definitely needs to be addressed.

PR #233, opened 2017-05-10
Inventory refactor and testing

dpastoor:
Hi Kyle,

I've dug through the inventory implementation and refactored a number of things to make it a bit more extensible and simplify the logic.

First, I refactored inven to be more consistent in what it returns, namely either it stops (when crumping) or it returns a vector of names for what is missing. This way, it can also be used to do things like tune other defaults or do other programmatic introspection downstream, rather than just giving a T/F plus a human consumable message. This also involved refactoring the underlying inven_report

Second, I refactored inventory to expose crump, however called it `.strict`. This is required by name, and gives the option for people to control the degree that they'd like to have the inventory warn them.

Third, I am returning the original model now, given inventory passes. This will allow inventory to be used in pipelines

```r
mod %>% 
    inventory(idata, everything()) %>% 
    idata_set(idata) %>% 
    mrgsim()
```

Finally, I added a number of tests. It is by no mean comprehensive at this point, but it gives some assurances that the various selection criteria are working as expected.  

I think there still needs to be a vignette/blog post expanding on this, but that may or may not need to be in scope for this PR.

### DIFF
--- a/R/inven.R
+++ b/R/inven.R
@@ -45,9 +45,9 @@ inventory <- function(x,obj,...) {
   dots <- lazyeval::lazy_dots(...)
   need <- select_vars_(names(param(x)),dots)
   if(length(need)==0) need <- names(param(x))
-  ans <- inven(x,obj,need,crump=FALSE) 
-  if(ans) message("Found all required parameters.")
-  return(invisible(ans))
+  inven(x,obj,need,crump=FALSE) 
+  message("Found all required parameters.")
+  return(x)
 }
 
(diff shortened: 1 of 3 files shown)


kylebmetrum:
@dpastoor Thanks for all of your work on this.  Much appreciated to have you step in now.  It's super busy for me until June.

dpastoor:
np - going to be stretching inventory for a couple ongoing projects so we'll see what else comes to light.

PR #234, opened 2017-05-12
Transition to dplyr6

kylebmetrum:


kylebmetrum:
None of this is needed.

PR #245, opened 2017-06-24
Rm select_ changes to inventory

kylebmetrum:
Current behavior:

``` r
library(rbenchmark)
library(dplyr)
```

``` r
library(mrgsolve)


mod <- mrgsolve:::house()

need <- c("CL", "SEX")

mydata <- data_frame(ID=1, CL=2, SEX=3)


inventory(mod,mydata)
```

    ## Warning: Missing parameters in 'mydata'
    ##  - VC
    ##  - KA
    ##  - F1
    ##  - WT
    ##  - WTCL
    ##  - WTVC
    ##  - SEXCL
    ##  - SEXVC
    ##  - KIN
    ##  - KOUT
    ##  - IC50

``` r
print(try(inventory(mod,mydata,everything())))
```

    ## [1] "Error : missing parameters in 'mydata'\n - VC\n - KA\n - F1\n - WT\n - WTCL\n - WTVC\n - SEXCL\n - SEXVC\n - KIN\n - KOUT\n - IC50\n"
    ## attr(,"class")
    ## [1] "try-error"
    ## attr(,"condition")
    ## <simpleError: missing parameters in 'mydata'
    ##  - VC
    ##  - KA
    ##  - F1
    ##  - WT
    ##  - WTCL
    ##  - WTVC
    ##  - SEXCL
    ##  - SEXVC
    ##  - KIN
    ##  - KOUT
    ##  - IC50>

``` r
print(try(inventory(mod,mydata,names(param(mod)))))
```

    ## [1] "Error : missing parameters in 'mydata'\n - VC\n - KA\n - F1\n - WT\n - WTCL\n - WTVC\n - SEXCL\n - SEXVC\n - KIN\n - KOUT\n - IC50\n"
    ## attr(,"class")
    ## [1] "try-error"
    ## attr(,"condition")
    ## <simpleError: missing parameters in 'mydata'
    ##  - VC
    ##  - KA
    ##  - F1
    ##  - WT
    ##  - WTCL
    ##  - WTVC
    ##  - SEXCL
    ##  - SEXVC
    ##  - KIN
    ##  - KOUT
    ##  - IC50>

``` r
inventory(mod,mydata,CL,SEX)
```

    ## Found all required parameters in 'mydata'.

``` r
inventory(mod,mydata,need)
```

    ## Found all required parameters in 'mydata'.

``` r
print(try(inventory(mod,mydata, c("VC",need))))
```

    ## [1] "Error : missing parameters in 'mydata'\n - VC\n"
    ## attr(,"class")
    ## [1] "try-error"
    ## attr(,"condition")
    ## <simpleError: missing parameters in 'mydata'
    ##  - VC>


R/inven.R
```diff
@@ -50,38 +50,38 @@ inven <- function(x,obj,need=NULL,crump=TRUE) {
...
   
-  need <- select_vars(names(param(x)), ...)
+  oname <- as.character(as.list(match.call())$obj)
+
+  need <- select_vars(names(param(x)),...)
   
-  if(!length(need)) {
-    return(x)
-  }
+  if(!length(need)) need <- names(param(x))
```
kylebmetrum: @dpastoor I think this is needed to make the "default" (nothing in `...`) parameter requirement all of the parameters in the model.

R/inven.R
```diff
@@ -31,35 +28,73 @@ inven <- function(x,obj,need="@all",crump=TRUE,...) {
...
   
-  if(!all(need %in% names(obj))) {
-    if(crump) inven_stop(x,obj,need)
-    return(inven_report(x,obj,need))
+  if(crump) {
+    inven_stop(obj,need)
   }
   
-  return(invisible(TRUE))
+  return(inven_report(obj, need))
```
dpastoor: It looks like below you changed the inven_report api to need oname as well

R/inven.R
```diff
@@ -31,35 +28,73 @@ inven <- function(x,obj,need="@all",crump=TRUE,...) {
     stop("need can only be parameter names", call.=FALSE) 
   }
   
-  if(!all(need %in% names(obj))) {
-    if(crump) inven_stop(x,obj,need)
-    return(inven_report(x,obj,need))
+  if(crump) {
+    inven_stop(obj,need)
```
dpastoor: what about oname?

R/inven.R
```diff
@@ -31,35 +28,73 @@ inven <- function(x,obj,need="@all",crump=TRUE,...) {
...
+inventory <- function(x,obj,..., .strict = FALSE) {
+  
+  oname <- as.character(as.list(match.call())$obj)
+
+  need <- select_vars(names(param(x)),...)
+  
+  if(!length(need)) {
+    need <- names(param(x))
+  } else {
+    if(missing(.strict)) .strict <- TRUE
```
dpastoor: won't strict never be missing since it is default of false

kylebmetrum:
We do need `inven`.  I will put it back, stop exporting, rename and make consistent with `inventory`.

```r
data_set.R:    inven(x,data,need) 
idata_set.R:  if(is.character(need)) inven(x,data,need)

```

R/inven.R
dpastoor: but that should just be handled by the default/or user provided values.


eg in the function spec just have function(,... , .strict = TRUE)

no need to check missingness, and reassign, if the user doesn't specify, then its true by default already.

That is the 'beauty' of forcing a full name match after ..., there is no real way for them to accidentally be specifying it if they didn't realize.

R/inven.R
```diff
@@ -48,13 +48,17 @@ inven <- function(x,obj,need=NULL,crump=TRUE) {
 #' }
 #' @return original mrgmod
 #' @export
-inventory <- function(x,obj,..., .strict = TRUE) {
+inventory <- function(x,obj,..., .strict = FALSE) {
```
dpastoor: based on your comment below, this should be changed back to true (which is my preference anyway)

dpastoor:
yes I thought you had inven as basically the internal logic/api, with inventory the user facing "cleaner" function with the nice dplyr api

kylebmetrum:
I'm putting it back but need to re-consider what is supposed to happen when the inventory 
is called from `data_set` or `idata_set`.  It seems like it should error if inventory isn't correct or just keep going if it isn't.  No warnings if no strict and no messages if everything checked out. 

Thoughts?

Either way, `data_set` and `idata_set` need to get updated to utilize the `select` stuff.

kylebmetrum:
@dpastoor  IC ... I thought you added some tests but I didn't see them.  They are in the master but not my `dev` branch so I messed that up somehow.  I'll bring your tests back in and update anything that needs.  Hopefully there is significant overlap and I can ditch the tests I wrote.

dpastoor:
yeah https://github.com/metrumresearchgroup/mrgsolve/blob/master/tests/testthat/test-inventory.R 

The inven_report internal function can be removed obviously since you refactored that out, but this should flex (a little) whether the existing top level api was still maintained or not

PR #328, opened 2018-03-04
Fix for parsing error in nmxml.R

callistosp:
Reviewing the source of the nmxml error, found that the tree structure was different than expected. Added an if/else statement to account for this variable structure (i.e. "output" element omitted from tree list, but all remaining elements present). Unsure why the structure was different, so this workaround may need to be implemented in a more general fashion in the future to account for additional variability.

### DIFF
--- a/R/nmxml.R
+++ b/R/nmxml.R
@@ -78,9 +78,11 @@ nmxml <- function(run=numeric(0), project=character(0),
   }
   
   tree <- xml2::as_list(xml2::read_xml(target))
-  tree <- ifelse(is.null(tree[["output"]]),
-                 tree$nonmem$problem$estimation,
-                 tree$output$nonmem$problem$estimation)
+  if(is.null(tree[["output"]])){
+    tree <- tree$nonmem$problem$estimation
+  } else {
+    tree <- tree$output$nonmem$problem$estimation
+  }
   
   th <- list()
   om <- matrix(0,0,0)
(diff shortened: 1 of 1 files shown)


kylebmetrum:
@callistosp Thanks.  Can you send up your `sessionInfo()` too?

callistosp:
@kylebmetrum sessionInfo() output:
R version 3.4.1 (2017-06-30)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] mrgsolve_0.8.10.9007

loaded via a namespace (and not attached):
 [1] compiler_3.4.1            magrittr_1.5             
 [3] assertthat_0.2.0          R6_2.2.2                 
 [5] tools_3.4.1               pillar_1.1.0             
 [7] bindrcpp_0.2              RcppArmadillo_0.8.400.0.0
 [9] glue_1.2.0                dplyr_0.7.4              
[11] tibble_1.4.2              Rcpp_0.12.15             
[13] xml2_1.1.1                pkgconfig_2.0.1          
[15] rlang_0.2.0               bindr_0.1

callistosp:
Thanks @kylebmetrum I updated the xml2 package on my computer and can confirm that I am no longer getting this error when I compile models using $NMXML. Your proposed implementation makes sense to me

Issue #518, opened 2019-09-27
Error in set record position with large(r) inputs


kylebaron:
Still some issues with sorting additional doses.  Non-random, based on the size of the input.  Also seen in current release. 

```
> mod <- modlib("pk2cmt", end = 168, delta = 1)
Loading model from cache.
> dose <- ev(amt = 100, ii = 24, addl = 6,cmt=2) #%>% realize_addl
> 
> idata <- expand.idata(CL = rep(1,1500))
> x <- chunk_by_row(idata,6)
> ans1 <- mrgsim_ei(mod, dose, x[[2]], output="df")
> ans2 <- mrgsim_ei(mod, dose, idata, output="df")
> ans3 <- filter(ans2, ID %in% ans1[["ID"]])
> filter(ans1, CENT != ans3$CENT) %>% head
[1] ID     time   EV1    CENT   PERIPH EV2    CP    
<0 rows> (or 0-length row.names)
> filter(ans3, CENT != ans1$CENT) %>% head
[1] ID     time   EV1    CENT   PERIPH EV2    CP    
<0 rows> (or 0-length row.names)
> 
> idata <- expand.idata(CL = rep(1,1501))
> x <- chunk_by_row(idata,6)
> ans1 <- mrgsim_ei(mod, dose, x[[2]], output="df")
> ans2 <- mrgsim_ei(mod, dose, idata, output="df")
> ans3 <- filter(ans2, ID %in% ans1[["ID"]])
> filter(ans1, CENT != ans3$CENT) %>% head
   ID time EV1     CENT   PERIPH EV2       CP
1 501   24   0 127.6592 16.37550   0 6.382958
2 501  120   0 151.0427 30.24358   0 7.552136
3 502   24   0 127.6592 16.37550   0 6.382958
4 502   48   0 140.6726 24.09336   0 7.033629
5 502   72   0 146.7997 27.72715   0 7.339983
6 502   96   0 149.6845 29.43805   0 7.484223
> filter(ans3, CENT != ans1$CENT) %>% head
   ID time EV1     CENT   PERIPH EV2       CP
1 501   24   0 27.65916 16.37550   0 1.382958
2 501  120   0 51.04271 30.24358   0 2.552136
3 502   24   0 27.65916 16.37550   0 1.382958
4 502   48   0 40.67258 24.09336   0 2.033629
5 502   72   0 46.79967 27.72715   0 2.339983
6 502   96   0 49.68447 29.43805   0 2.484223
[...]

PR #519, opened 2019-09-28
Vars

kylebaron:
This started as experimenting with openmp.  It worked, but decided to drop it in the end. 

- Kept thread-safe model specification
- Kept cxxlsoda solver
- Switched to C++11 shared pointers rather than boost; keeping boost tokenizer for now
- FIxes bug related to record sorting with large idata sets #518 
- Adds explicit tests for recsort (tests alongside stated behavior)
- Refactored qsim and mrgsim_q a bit; takes capture only by default


kylebaron:
See #529 This reverts the thread-safe model specification.

PR #537, opened 2019-10-04
Create validation test directory in inst

kylebaron:
- Run all tests with Makefile
- Just two tests right now; will add more as needed


kylebaron:
wrong branch

Issue #533, opened 2019-10-04
[F] Refactor control of advance to ss


kylebaron:
# Summary

New arguments for mrgsim to control how the PK system is advanced to steady state.  The basic process is the same for bolus, intermittent infusion, or ss infusion, but these parameters can expose some control over determining ss.

- `ss_n` max number of iterations (doses to administer) to try
- `ss_fixed` assume ss is achieved after `ss_n` iterations
- a warning is generated if the PK system does not advance to steady state with specified tolerance within specified maximum number of iterations unless `ss_fixed` is true
- ss is achieved when the difference between the current trough and the previous trough is less than the local error estimate based on model atol and rtol

See PR #557

See https://github.com/metrumresearchgroup/mrgsolve/commit/450d97ce14334aee0f548754067c793a2228e8dd

From: http://www.netlib.no/netlib/alliant/ode/prog/lsoda.f
```
c rtol   = relative tolerance parameter (scalar).
c atol   = absolute tolerance parameter (scalar or array).
c          the estimated local error in y(i) will be controlled so as
c          to be less than
c             ewt(i) = rtol*abs(y(i)) + atol     if itol = 1, or
c             ewt(i) = rtol*abs(y(i)) + atol(i)  if itol = 2.
c          thus the local error test passes if, in each component,
c          either the absolute error is less than atol (or atol(i)),
c          or the relative error is less than rtol.
```

# Tests
- tests/testthat/test-ss.R
  - ss_n and ss_fixed issue-533
- inst/maintenance/unit/test-pk.R
  - one-compartment, bolus, ss
  - one-compartment, oral, first, ss
  - two-compartment, bolus, ss
  - two-compartment, bolus, first, ss
  - two-compartment, infusion, ss

dpastoor:
LGTM - though there might be some user ergonomic changes that make sense around tolerances that have been discussed internally. Should create a new issue

PR #557, opened 2019-10-07
Refactor/ss control

kylebaron:
Updates #533


src/odeproblem.cpp
```diff
@@ -87,7 +87,7 @@ odeproblem::odeproblem(Rcpp::NumericVector param,
   d.amt = 0;
   
   Do_Init_Calc = true;
-  ss_tol = 1e-12;
+  ss_fixed = false;
   ss_n = 1000;
```
yizhang-yiz: Why is `ss_n` in this constructor different from that in the header? Here we have 1000, in the header it's 500. Is there anywhere `odeproblem` will be constructed using the default constructor?

EDIT: if this is going to be always fixed with a default value, isn't previously defined `MRGSOLVE_MAX_SS_ITER` for this purpose?

src/datarecord.cpp
```diff
@@ -233,26 +230,22 @@ void datarecord::steady_bolus(odeproblem* prob, LSODA& solver) {
   }
   
   prob->rate_reset();
-  double N_SS = prob->ss_n;  
-  double CRIT_DIFF_SS = prob->ss_tol;
+  bool warn = !prob->ss_fixed;
+  int N_SS = prob->ss_n;  
+  int min_check = (N_SS < 10 || prob->ss_fixed) ? N_SS : 10;
+
```
yizhang-yiz: If I understand it correctly, this is essentially what `ss_fixed` is for: to control the number of dosing/ODE solutions, and with its `false` default we do maximum 10 doses. Does this number 10 cover majority of cases?

src/datarecord.cpp
```diff
@@ -233,26 +230,22 @@ void datarecord::steady_bolus(odeproblem* prob, LSODA& solver) {
   }
   
   prob->rate_reset();
-  double N_SS = prob->ss_n;  
-  double CRIT_DIFF_SS = prob->ss_tol;
+  bool warn = !prob->ss_fixed;
```
yizhang-yiz: why no warning with `ss_fixed=true`?

inst/include/odeproblem.h
```diff
@@ -201,8 +201,8 @@ class odeproblem {
   int Npar;
   int Neq;
   int Istate; ///< istate value
-  double ss_tol;
-  double ss_n;
+  bool ss_fixed = false;
+  int ss_n = 500;
```
yizhang-yiz: Will `odeproblem` be ever constructed with these default values?

src/LSODA.h
```diff
@@ -51,8 +51,8 @@ class LSODA
     int iopt; 
     int jt;
     int itask;
-    double Atol; 
-    double Rtol;
+    double Atol=1e-8; 
+    double Rtol=1e-8;
```
yizhang-yiz: will `LSODA` class be ever constructed with these default values? when will these values be in effect? Is that scenario in unit test?

src/odeproblem.cpp
kylebaron: - Removed the default constructor (wasn't used anywhere)
- Set initialize ss_n to 500 in the (one) constructor
- removed `MRGSOLVE_MAX_SS_ITER` (legacy)

src/datarecord.cpp
kylebaron: - There might be cases where the system won't go to steady state
- If the user gets into this situation, they will get warned
- `ss_fixed` is a way to let the user say "just give 100 doses and call that steady state" and 
the warning goes away at this point

inst/include/odeproblem.h
kylebaron: These default values were removed.

src/LSODA.h
kylebaron: - Moved all initialization to the constructor
- Removed default constructor

src/datarecord.cpp
kylebaron: - My thinking on this was to run 10 iterations up front before starting to check
- Some scenarios would get there before 10 but it could be all over the place
- Maybe it has limited usefulness

R/mrgsolve.R
```diff
@@ -450,9 +450,15 @@ mrgsim_nid <- function(x, nid, events = ev(), ...) {
...
 ##' @param ss_n maximum number of iterations for determining steady state for 
-##' the PK system
+##' the PK system; a warning will be issued if steady state is not achieved 
+##' within \code{ss_n} iterations when \code{ss_fixed} is \code{TRUE}
+##' @param ss_fixed if \code{TRUE}, then a fixed number of iterations are 
+##' performed (determined by \code{ss_n}) for determining steady state without
+##' regard to model tolerances; to silence warnings related to steady state, 
+##' set \code{ss_fixed} to \code{TRUE} and set \code{ss_n} as the number of 
+##' iterations to advance the system for steady state determination
+##'
```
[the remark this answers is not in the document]
kylebaron: Added documentation for `FALSE`

kylebaron:
It looks like there is a new RcppArmadillo out today and it is not getting installed on Travis (or rhub); but checks are passing locally.

```

Version: | 0.9.800.1.0
-- | --
Depends: | R (≥ 3.3.0)
Imports: | Rcpp (≥ 0.11.0), stats, utils, methods
LinkingTo: | Rcpp
Suggests: | tinytest, Matrix, pkgKitten, reticulate, rmarkdown, knitr, pinp, slam
Published: | 2019-10-09

```

kylebaron:
The build eventually ran:
https://travis-ci.org/metrumresearchgroup/mrgsolve/builds/595668392

PR #569, opened 2019-10-12
feature: adds ID number to ss fail message

kylebaron:
I just thought of this;  not required to function, but I'm sure people will ask for it and it will be useful when simulating a population. 

Ok to dismiss if you'd rather just bring it in at the next release.

### DIFF
--- a/src/datarecord.cpp
+++ b/src/datarecord.cpp
@@ -277,8 +277,8 @@ void datarecord::steady_bolus(odeproblem* prob, LSODA& solver) {
   if((!made_it) && warn) {
     Rcpp::warning(
       tfm::format(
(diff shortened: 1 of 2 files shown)


dpastoor:
at this point, the release should be under a code freeze during acceptance, so I think it will be best to push this off into the next release. That said, we can do a low risk immediate follow up release that just contains this. 

From a validation perspective, we should do them one after the other, on the same day, monday, such that CRAN could still get this 'version' while this addition doesn't trigger any reassessment or formal impact analysis which would be required to add anything post-code-freeze.

What do you think about that instead?

kylebaron:
SOunds good.

PR #635, opened 2020-02-12
Release/0.10.1

kylebaron:
- [ ] bumped package version to 0.10.1
- [ ] NEWS.md is updated, with version number
- [ ] NONMEM tests [test/0.10.1](https://github.com/mrgsolve/nmtests/pull/1)


R/update.R
```diff
@@ -308,8 +309,77 @@ setMethod("update", "parameter_list", function(object,.y,...) {
...
+  i <- as.list(init(data))
+  up <- list(
+    start = data@start, end = data@end, delta = data@delta, 
+    atol = data@atol, rtol = data@rtol, hmax = data@hmax,
+    maxsteps = data@maxsteps
+  )
+  up <- up[which(!(names(up) %in% c(names(p),names(i))))]
+  parent <- parent.frame()
+  env <- list2env(c(p,i,up),parent = parent)
+  eval(substitute(expr),env)
```
dpastoor: whats the point of this eval here, it doesn't get saved anywhere? ... or does this evaluate in the environment? I'm not familiar with this code, but if I had to guess it would eval in the env, but just want to confirm.

inst/base/databox_cpp.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2013 - 2019  Metrum Research Group, LLC
+// Copyright (C) 2013 - 2019  Metrum Research Group
```
dpastoor: 2020

tests/testthat/test-mrgsims.R
```diff
@@ -31,3 +31,10 @@ test_that("mrgsims class", {
   expect_is(as.list(out), "mrgsims_list")
 })
 
+test_that("plot from character", {
+  out <- mrgsolve:::house() %>% mrgsim()
```
dpastoor: exported house now

tests/testthat/test-mrgmod.R
```diff
@@ -26,18 +26,22 @@ context("test-mrgmod")
...
+  expect_equal(mod[["RESP"]],50)
+  expect_equal(mod$end,120)
+  expect_equal(mod[["end"]],mod@end)
   expect_error(mod$kylebaron) 
-  expect_error(mod[["kylebaron"]])
+  expect_error(mod[["kylebaron"]],regexp = "not found or not extractable")
   expect_is(as.list(mod), "list")
   expect_output(summary(mod), "Model: housemodel")
   expect_true(mrgsolve:::valid.mrgmod(mod))
   expect_true(all.equal(mod, mrgsolve:::house()))
```
dpastoor: exported house

R/update.R
```diff
@@ -308,8 +309,77 @@ setMethod("update", "parameter_list", function(object,.y,...) {
...
+#' Other model object slots that can be updated: `start`, `end`, `delta`, 
+#' `add`, `rtol`, `atol`, `hmax`, `maxsteps`.  These are include for convenience, 
+#' but we expect that most of the time these will get updated through the 
+#' update method.
+#' 
+#' @param data an object with class mrgmod
+#' @param expr expressions evaluated in an environment containing various model
+#' object components, including parameters, initial conditions, and others 
+#' (see details)
+#' @param ... not used
```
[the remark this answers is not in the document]
kylebaron: I think it's required here:

>A method must have all the arguments of the generic, including … if the generic does.

https://cran.r-project.org/doc/manuals/R-exts.html#Generic-functions-and-methods

Will take it out if it can be taken out.

R/model_include.R
```diff
@@ -67,14 +67,14 @@ make_clink <- function(x,clink) {
   paste(paste0("-I\"",unique(link), "\""),collapse=" ")
 }
 
-set_clink <- function(x,clink=NULL) {
+set_clink <- function(x,clink=NULL,...) {
```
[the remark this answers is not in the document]
kylebaron: Arguments are getting passed to multiple functions from the calling function.  The dots absorb unused arguments.  What is the right thing to do?  Just get explicit in the calling function and receive all arguments and the pass only the ones needed to functions downstream?

R/update.R
kylebaron: I think it evals in `env`;  I was following `base:::within.list`. 
```r
function (data, expr, keepAttrs = TRUE, ...) 
{
    parent <- parent.frame()
    e <- evalq(environment(), data, parent)
    eval(substitute(expr), e)
    if (keepAttrs) {
        l <- as.list(e, all.names = TRUE)
        del <- setdiff(names(data), (nl <- names(l)))
        data[nl] <- l
        data[del] <- NULL
        data
    }
    else {
        as.list(e, all.names = TRUE)
    }
}
``` 

I'm not totally sure about `evalq` then `eval`.  But I think setting `parent` in `list2env` does something along the same lines?  I can access a variable in the parent frame to use in `expr`

```r
x <- 2
within(mod, {
  CL = CL*x
})
```

kylebaron:
@dpastoor thanks so much for the review and comments;  I do think the dots need to stay with the `within` method.  Something needs to get re-factored for the `set_clink` thing and open to suggestions for what to do there.  None of it is user facing, so I guess I could just get explicit about arguments and passing along without needing awkward explanations in the documentation.  But will wait for advice. 

I removed `mrgsolve:::house()` from all examples and the primary test bed.  I held off on the secondary location.  Just want to work on refactoring tests so that they are a bit more manageable.  So going to hold off there.  

I believe comments are addressed  but please throw something at me if I missed something.

NEWS.md
```diff
@@ -1,3 +1,22 @@
+# mrgsolve 0.10.1
+
```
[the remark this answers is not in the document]
kylebaron: NEWS was updated.

dpastoor:
All makes sense to me, looks ready to go!

R/update.R
```diff
@@ -308,8 +309,77 @@ setMethod("update", "parameter_list", function(object,.y,...) {
...
+#' 
+#' @examples
+#' mod <- mrgsolve:::house()
```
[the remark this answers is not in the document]
kylebaron: Should be using `mrgsolve::house` exclusively in the examples and main test grouping.

inst/base/databox_cpp.h
kylebaron: Updated this and a bunch of others.

tests/testthat/test-mrgsims.R
kylebaron: Addressed.

tests/testthat/test-mrgmod.R
kylebaron: Addressed.

PR #637, opened 2020-02-13
change the flag in CXX flag test so test behaves same for gcc

kylebaron:
In test `inst/validation/test-validation.R/PKG_CXXFLAGS is set issue-603`:

- Update test so that test behavior is consistent on gcc as well as clang
- No change to package functionality or behavior; everything continues to behave properly and as expected; only tweaking test so that the test can accommodate different compilers

I will write this up as a deviation.


kylebaron:
I am merging this PR; but it was discussed and approved by Tech-Solutions personnel.

PR #662, opened 2020-03-23
Optionally read the ext file from NMXML

kylebaron:
- Required to pass `file`
- Use `data.table::fread` if available; otherwise, use `read.table`
- Error if cannot find the ext file
- Error if we filter on the estimates row, but nothing is there

Super important note: user will always get a single omega matrix and a single sigma matrix when using this.


kylebaron:
@timwaterhouse Looking for thought / feedback at this point.

kylebaron:
Awesome, @timwaterhouse .  Thanks for doing the benchmarks too!  That really helps.  I honestly wasn't sure how that would turn out but helps support this direction a bit. 

It looks like the xml / ext approaches will result in equivalent results **except** for the structure of OMEGA or SIGMA.

So if NONMEM was

```
$OMEGA BLOCK(2)
1
0.01 2

$OMEGA
1
```
Then the mrgsolve implementation would always be like this if you read from the ext file

```
$OMEGA @block
1
0.01 2
0 0 1
```

But the mrgsolve representation would match that in NONMEM if you read from xml.  There would be no functional difference in the way the model runs, but there would be difference in how you update $OMEGA in the mrgsolve version.

Wondering your thoughts on this;

1. Big deal or not a big deal?
1. Anywhere else we could quickly find the structure? If we could, we could work backwards to get the same representation.

kylebaron:
I guess another question:

We can always change what happens under the hood.  Is `read_ext = TRUE` going to be a stable way to request this?   We can always add more options later, but does this seem sensible basic way to tell mrgsolve what to do?

timwaterhouse:
> Big deal or not a big deal?

I guess I'd vote for not a big deal?  I don't foresee that I'd be switching between using .ext and .xml files a lot, so I think once the model is set up I'd know how to update $OMEGA.

> Anywhere else we could quickly find the structure? If we could, we could work backwards to get the same representation.

Maybe check off-diagonals for zeros?  Not sure if that would be totally fool-proof.  There's also the block structure in the .lst file (`0OMEGA HAS BLOCK FORM`), but I'm guessing it would be a pain trying to parse that.

> We can always change what happens under the hood. Is read_ext = TRUE going to be a stable way to request this? We can always add more options later, but does this seem sensible basic way to tell mrgsolve what to do?

I did wonder if a totally new block like `$NMEXT` might make more sense.  Seems a bit odd to use `$NMXML` when we're not using the .xml file.

kylebaron:
The nmext version is available here
https://s3.amazonaws.com/mpn.metworx.dev/releases/mrgsolve/0.10.1.9006-2



``` r
library(tidyverse)
library(mrgsolve)
#> 
#> 
#>     filter


library(mrgsolve)

code <- '
$NMEXT path = system.file("nonmem", "1005", "1005.ext", package = "mrgsolve")
'

mod <- mcode("foo", code)
#> Building foo ...
#> done.

param(mod)
#> 
#>  Model parameters (N=7):
#>  name   value  . name   value
#>  THETA1 9.51   | THETA5 113  
#>  THETA2 22.8   | THETA6 1.02 
#>  THETA3 0.0714 | THETA7 1.19 
#>  THETA4 3.47   | .      .
revar(mod)
#> $omega
#> $...
#>           [,1]       [,2]       [,3]
#> 1:   0.2138790  0.1207700 -0.0116278
#> 2:   0.1207700  0.0945105 -0.0372064
#> 3:  -0.0116278 -0.0372064  0.0465631
#> 
#> 
#> $sigma
#> $...
#>          [,1]     [,2]
#> 1:  0.0491707 0.000000
#> 2:  0.0000000 0.201769
```

<sup>Created on 2020-04-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

timwaterhouse:
Is that URL for this version correct? I tried installing from there and it doesn't seem to recognise `$NMEXT`.

``` r
  packageVersion("mrgsolve")
#> [1] '0.10.1.9006.2'
  library(mrgsolve)
#> 
#> 
#>     filter
  
  code <- '
$NMEXT path = system.file("nonmem", "1005", "1005.ext", package = "mrgsolve")
'
  
  mod <- mcode("foo", code)
#> Warning: Invalid blocks found: NMEXT.
#> Error: No valid blocks found.
```

<sup>Created on 2020-04-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

kylebaron:
Ack;  I'm sorry, @timwaterhouse it's the `dash-3` version.  I was thinking about correcting the link as I was copying and pasting, but didn't actually do it.  :(
 
https://s3.amazonaws.com/mpn.metworx.dev/releases/mrgsolve/0.10.1.9006-3

timwaterhouse:
Looks great, thanks Kyle! This works fine for my test model.

Issue #690, opened 2020-06-16
Feature Request (revisited): Assign Omegas and Sigmas by Name


billdenney:
@kylebaron and @dpastoor, I hope that you're both doing well.

As is often the case for me, when I have an issue once, I have it repeatedly.  In this case, I have a set of models where during development the ordering of omega values in the NONMEM model are changing due to checking different block structures, and so it is more difficult to rapidly put those models into mrgsolve.  (Is IIVCL in omat 1 or 2?  Which order do I need to put in IIVCL vs IIVVC vs IIVVMAX?)

In #461, I requested the ability to assign an entire matrix simultaneously, and I acknowledge that seems complex in several ways.  Here, I'm asking for the ability to assign into the omega and sigma blocks with a 1x1 named matrix (where I assume it would always assign symmetrically).  Overall, I'm thinking that this would look like an extra step in `omat(x, ...)` for the matrix signature adding something like an `assign_by_name` argument to ensure that the feature is only used intentionally and that it would not cause a regression for previous models that may not expect the name to be used during assignment.

I recognize that I'm rehashing an old question (and, I apologize about that).  But as it is an issue that I have had several times, I'm trying to find the best solution for my issue.

If this isn't viewed as a good fit, then I'm curious what your recommended workflow is to ensure accurate assignment into the omega and sigma matrices when the source parameter order may change?

kylebaron:
@billdenney is this what you want to do?  I was thinking this was more complex if you had to track off-diagonals.  But as you state the problem, there isn't any guarantee that any two will have any covariance, so all you can do is update the diagonals


``` r
library(tidyverse)
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '
$OMEGA @labels a, b, c, d @name first
1 2 3 4

$OMEGA @labels aaa, bbb , ccc @name second
111
222
333


$OMEGA @labels aa, bb @name third
11
22



'
mod <- mcode("ma", code)
#> Building ma ...
#> done.

assign_by_name <- function(mod,...,.type = c("omat", "smat")) {
  l <- list(...)
  stopifnot(rlang::is_named(l))
  fun <- getFromNamespace(match.arg(.type),"mrgsolve") 
  ml <- fun(mod)
  d <- dim(ml)
  x <- as.matrix(ml)
  lab <- unlist(labels(ml))
  dimnames(x) <- list(lab,lab)
[...]

PR #691, opened 2020-06-16
Add names to all omega and sigma matrices and the matrix-creation functions

### LINKED ISSUE #463: Feature Request: Assign Names to Omega and Sigma Matrices
This is tightly related to #461, but it given the way that the conversation is going there, it feels like a separate issue.

When extracting the omega and sigma matrices, can the rows and columns be named to assist the user in assigning values to the correct matrix location?  Specifically, if annotated with names, can the rownames and colnames be assigned with the annotated names?  (If not annotated, the matrix would not have names.)

billdenney:
This effectively fixes #463, and it is related to #690.

The goal of this PR is to allow name addition to all omega and sigma matrices along with the `bmat`, `dmat`, and `cmat` methods.  Given the discussion in #463, I'm not sure if this PR will be accepted, so I've not yet gone to the effort of updating or writing tests for it (and, it may break other tests because running tests locally was not immediately possible with `devtools::test()` because of missing directories).  If acceptable and helpful, I will write the tests.

Also, I'm not certain if the change that I made to `SUPERMATRIX()` may break other things.  I assume that would show up in testing.

Many testing errors looked like the following.  I don't think that I'm causing new errors in areas unrelated to the matrix changes that I made, but I'm not certain:

```
test-nmxml.R:47: error: (unknown)
'C:/git/Not_my_repositories/mrgsolve/nonmem/1005/1005.xml' does not exist.
Backtrace:
  1. mrgsolve::mcode("test6", code, compile = FALSE) C:\git\Not_my_repositories\mrgsolve/tests/testthat/test-nmxml.R:47:0
  2. mrgsolve::mread(...) R/mcode.R:56:2
  3. base::lapply(spec, handle_spec_block, env = mread.env) R/mread.R:229:2
  5. mrgsolve::handle_spec_block.specNMXML(X[[i]], ...) R/handle_spec_block.R:26:21
  6. mrgsolve::parseNMXML(dump_opts(x), ...) R/handle_spec_block.R:355:2
 10. xml2:::read_xml.character(target)
 11. xml2:::path_to_connection(x)
 12. xml2:::check_path(path)
```


billdenney:
There was also a change that appeared to be introduced by running `devtools::document()`.  I've made it here for completeness, but it is not related to what I changed.

kylebaron:
@billdenney - 

I'd rather not mess with this I kept the names separate by design ... it simplifies things a bit, IMO. 

One thing I can do: we can improve the `as.matrix` function to add dimnames to the output so you'd have that to work with.  You can always coerece it to a list of matrices and it would be really easy to name that too.  That way, the internals can stay as they are without having to worry about names ... and you can get the output with names attached if you wish. 

Also could do this on the input: if a matrix is fully named, we could pick those up as labels.

Would that meet your needs?

billdenney:
@kylebaron , I understand the goal of keeping things simpler.  I think that having matrix names when extracted (as you say with the `as.matrix()` methods) would go a long way toward what I'm looking for as I could use those names for alignment.  Thanks.

Issue #702, opened 2020-06-30
Feature: return time after dose in a specific compartment


kylebaron:
# Summary

As a user, I want to be able to calculate the time after a dose to a specific compartment

# Tests

- inst/maintenance/unit/test-tad-cmt.R
  - tad-cmt matches tad-no-cmt
  - check against tad
  - tad-cmt with default initialization

billdenney:
I'm commenting to follow the issue.

kylebaron:
@billdenney I have some code worked out for this; would you be willing to give it a spin?

billdenney:
Happily!  How do I test it.

kylebaron:
Thanks.  Let me get the doxygen docs updated; I think that will help.  SHould be close; will let you know.

kylebaron:
``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
code <- '
$cmt foo bar

$param CL = 10, V = 20, KA = 1

$plugin tad

$pkmodel depot = TRUE

$main
static mrg::tadose x(2);
static mrg::tadose y(1);
capture tadx = x.tad(self);
capture tady = y.tad(self);

$table
capture tados = self.tad();
'

mod <- mcode("adfa", code, end = 48)
#> Building adfa ...
#> done.

dose <- c(
  ev(amt = 100, time = 1, cmt = 2, ii = 4, addl = 9),
  ev(amt = 300, time = 4, cmt = 1, ii = 8, addl = 4)
)

a <- mrgsim(mod,dose)
plot(a)
```

![](https://i.imgur.com/bfhy4lj.png)

<sup>Created on 2020-09-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

kylebaron:
https://mrgsolve.github.io/doxygen/classmrgsolve_1_1tadose.html

kylebaron:
The was forked from issue #700

PR #746, opened 2020-09-18
Feature/plugin tad

kylebaron:
#702 

As a user, I want to be able to calculate the time after dose for specific dosing compartments. 

This functionality is exposed by invoking the `tad` plugin (`$PLUGIN tad`)

This implements a `tadose` class whose objects can track specific compartment numbers.  To track time after dose, create an object and register the compartment number to track.  When the `.tad()` method is called, it will return the time after dose.  Time after dose is -1 until the first dose is administered.


kylebaron:
@yizhang-yiz does technical review, @timwaterhouse does functional review.  Please respond here about when you can get this review done.

yizhang-yiz:
I can have it done by end of day.

timwaterhouse (approved):
Functionality looks good.

inst/base/mrgsolve-tad.h
```diff
@@ -0,0 +1,109 @@
...
+  double tad(databox& self);
+  void reset();
+};
+/**
+ * Constructor.
+ * 
+ * @param cmt_ compartment number
+ * 
+ */
+tadose::tadose(int cmt_) {
```
yizhang-yiz, line 65: In general it's not a good idea to leave things uninitialized. Above all it simply makes it hard to debug. So for `cmt` we can give it a large number, such as `std::numeric_limits<int>`. For `had_dose` we need to set it to `false`.

inst/base/mrgsolve-tad.h
```diff
@@ -0,0 +1,109 @@
...
+ * 
+ * @param self the model `databox` object
+ * 
+ * @return the time after dose or `-1` if no dose has been given
+ * 
+ */
+double tadose::tad(databox& self) {
+  if(self.newind <= 1) this->reset();
+  if(self.evid==0) {
+    return had_dose ? self.time - told : -1.0;
```
yizhang-yiz, line 90: Is there a meaning of `time=-1.0`? If so we need use something else. 

This line does not necessarily require any action.

yizhang-yiz (approved):
Everything's good all around. Only a minor suggestion.

inst/base/mrgsolve-tad.h
kylebaron: I should just include a comment like this in a PR review template so you don't have to type it out every single review :). 

1. Providing default values for both `cmt` and `had_dose`
1. Including default initialization in the test model

inst/base/mrgsolve-tad.h
kylebaron: I use -1 in other time after dose functionality; so that has come to mean there hasn't been a dose yet.

yizhang-yiz (approved):

Issue #755, opened 2020-11-05
deprecate mrgsolve:::render


kylebaron:
deprecate warn for now

PR #756, opened 2020-11-05
deprecate render inside Rmd

kylebaron:
Un-exported functionality that we will no longer support. #755


timwaterhouse (approved):
I'm fine with deprecating . I'm not sure how this typically works. Do you add this warning for a few releases and then eventually remove functionality?

kylebaron:
@timwaterhouse Yeah; just warn for now (once every 8 hours) so people know its coming. Then change it to error.

PR #757, opened 2020-11-06
[wip] [do not merge] refactor pipe functions for recover, outvars, etc

kylebaron:
From #697 

I'm not crazy about this, but `utils::recover()` already exists

I wanted to add pipe functions for setting `outvars` and `recover`

What I ended up with:

- we already have `carry_out()`
- add `recover_out()` which just sets `recover` in `mrgsim()` (like `mod %>% recover(WT)` but it's `mod %>% recover_out(WT)`
- add `vars_out()` which updates outvars (like `mod %>% update(outvars = c("Var1", "Var2"))` but it's `mod %>% vars_out(Var1, Var2)`

So giving up on trying to match the argument name (like `recover`) and using these `_out` variants instead. 

One thing we could do is change the name of the argument (something else other than `recover`).  I'm not against that but thought this proposal would be less offensive.  

@timwaterhouse What do you think?


kylebaron:
maybe @dpastoor comment too; I'm not crazy about what I came up with; just want to look at alternatives before trying to export a function called `recover()`.

timwaterhouse (approved):
I'm good with adding these `_out` functions rather changing arguments.

dpastoor:
Purely from a naming convention my immediate reaction is that recover is somewhat of a keyword or concept for fixing after failure. In fact when I saw this issue I thought this might have even been a helper to recover in the case of internal model failures during a simulation. Have we looked at other packages in the tidy models ecosystem or broom or the like to see if they have any such naming conventions? I believe there are things like augment but that doesn't immediately bring any warm fuzzies either.

Given my nuanced knowledge of MRG solve is definitely atrophying without looking do we already have a function capture? That would align with the idea of capturing outputs.

To your original question though and this current implementation recover_out if I was doing it myself i would give as a full name recover_outputs I always at least thought of carry out as literally being like the action of carrying something out of the model internals, not necessarily out as a shorthand for outputs. 

 recover_out doesn't have the same connotation however I don't take issue with it either

kylebaron:
- `recover` is currently the name of the argument; the goal was to add a function that manipulates that in pipe fashion
- it selects inputs that you want carried (copied) into the output without working with those data inside the model
- it is similar to `carry_out` but can handle any type
- `recover` works in a different way than `carry_out` and depends on `carry_out` functionality
- I don't know if anything comparable exists in the tidy world
- one option is to change the name of the argument if it is that hard to deal with 
- if we need to step back and comprehensively reconsider things, we can do that; there certainly is precedent for that sort of thing

kylebaron:
The original idea was:  `to save from loss and restore to usefulness : reclaim`; i don't doubt that this isn't the first thing that comes to mind for people who are new to the party

kylebaron:
This was removed from the next release milestone.  Will take it up later.

Issue #745, opened 2020-09-15
Reset solver when copying parameters from data


kylebaron:
# Summary

As a user, I want to be able to trigger hard ODE solver reset at when parameters are copied from the data
# Tests

kylebaron:
See example posted by @FelicienLL in #741

kylebaron:
This turns out to be very slow, depending on the problem.  Let's hold on this for the moment.

FelicienLL:
Ok thanks, nevermind. I think I faced an example that didn't pass the "evid=2 or hmax=0.01" strategy the other day. I was considering to simplify my model and data into a reprex to ensure it's related to time-varying cov. If so, would you still be interested in me posting it here to turn it into a test ? 
Thanks again for trying.

FelicienLL:
Hi, 
Just reporting an other example here in case you want to investigate this problem again one day. Here, predictions can differ from NONMEM if there is a lag time, and the covariate value changes in an obervation record after drug depot. Forcing solver to reset with ```evid =2``` does not correct it (while it worked in https://github.com/metrumresearchgroup/mrgsolve/issues/741#issuecomment-692597206 ). Fixing  ```hmax = 0.01``` partially fixed it (up to 5 digits in this example, is it considered satisfying ?). There are no more discrepancies between NONMEM and mrgsolve if models are coded in closed form nor if there is no lag time anymore (data not shown, but I can if of interest).

Thanks
Félicien

``` r
####  REPREX #####
library(tidyverse)
#> Warning: le package 'tibble' a été compilé avec la version R 4.0.3
library(mrgsolve)
#> 
#> Attachement du package : 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
packageVersion("mrgsolve")
#> [1] ‘0.10.4.9010’

setwd("C:/Documents and Settings/le_louedec/Documents/REPREX/001/")
options(pillar.sigfig = 6)

#DATA#

data <- tibble(ID = 1, time = c(0, 24, 48), evid = 1, cmt = 1, amt = 1000, mdv = 1,  COV = 1) %>%
  add_row(ID = 1, time = c(49, 50), evid = 0, cmt = 2, amt = 0, mdv = 1, COV = 2) %>%
  bind_rows(mutate(., ID = 2, COV = 1), mutate(., ID = 3, COV = 2)) %>%
  mutate(DV = NA_real_)
print(data)
#> # A tibble: 15 x 8
[...]

PR #761, opened 2020-11-11
reset solver when copying parameters

kylebaron:
Issue #745

- reset solver any time parameters are copied from the data
- see the issue; there were solver overshoot issues when time varying covariates changed parameter values; this is fixed by re-initializing solver whenever we copy parameters

### DIFF
--- a/src/dataobject.cpp
+++ b/src/dataobject.cpp
@@ -202,6 +202,7 @@ void dataobject::copy_next_parameters(int id_n, bool from_data, int this_row,
   if(done_copying) return;
   if(from_data) {
     copy_parameters(this_row, prob);
+    prob->lsoda_init();
     if(this_row >= Endrow.at(id_n)) {
       done_copying = true;  
       return;
@@ -212,6 +213,7 @@ void dataobject::copy_next_parameters(int id_n, bool from_data, int this_row,
   if((next_copy_row != last_copy_row) && (next_copy_row <= Endrow.at(id_n))) {
     copy_parameters(next_copy_row, prob); 
     last_copy_row = next_copy_row;
+    prob->lsoda_init();
   }
 }
 
--- a/src/dataobject.cpp
+++ b/src/dataobject.cpp
@@ -54,7 +54,7 @@ dataobject::dataobject(Rcpp::NumericMatrix _data,
   
   col.resize(8,0);
   
-  any_copy = parnames.size() > 0;
+  any_copy = par_from.size() > 0;
   done_copying = false;
   last_copy_row = -1;
   next_copy_row = 0;
@@ -88,7 +88,7 @@ dataobject::dataobject(Rcpp::NumericMatrix _data,
   
   col.resize(8,0);
   
-  any_copy = parnames.size() > 0;
+  any_copy = par_from.size() > 0;
   done_copying = false;
   last_copy_row = -1;
   next_copy_row = 0;
@@ -186,10 +186,10 @@ void dataobject::idata_row() {
 
 void dataobject::copy_parameters(int this_row, odeproblem* prob) {
   size_t n = par_from.size();
-  for(size_t i=0; i < n; ++i) {
+  for(size_t i = 0; i < n; ++i) {
     prob->param(par_to[i],Data(this_row,par_from[i]));
   }
-  prob->lsoda_init();
+  if(n > 0) prob->lsoda_init();
 }
 
 void dataobject::next_id(int id_n) {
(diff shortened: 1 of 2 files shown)


yizhang-yiz (commented):
Can we just call `prob->lsoda_init` inside `copy_parameters` or there are other scenarios when we need copy param but don't want to reinitialize `prob`? 

When `istate=1` is set iby default lsoda will redo time step adaptation, this involves overhead but I guess it's hard for us to know what types of covariant change require re-init. What you've tried in the other thread with small hmax essentially forces solver to climb a possible discontinuity, which could be time consuming and inaccurate.

kylebaron:
@yizhang-yiz Yeah, I was trying to avoid doing this if possible b/c it does slow things down.  There has always been some inaccuracy around this, but honestly it seems to be worse since I switched to that C++ implementation of dlsoda.  It's hard to say but seems that way.  I'm not sure if there is anything we could look into in the solver or some way we could do what we need for the reset without giving up time step adaptation.  Is this worth looking into?

yizhang-yiz (approved):

yizhang-yiz:
>  Is this worth looking into?

In general yes but that'll depend on how much the two solvers diverges and the degree of slowdown. Since I can't think of any way to detect parameters that would cause unnecessary reinitialization, the current design makes sense to me.

src/dataobject.cpp
```diff
@@ -186,10 +186,10 @@ void dataobject::idata_row() {
 
 void dataobject::copy_parameters(int this_row, odeproblem* prob) {
   size_t n = par_from.size();
-  for(size_t i=0; i < n; ++i) {
+  for(size_t i = 0; i < n; ++i) {
     prob->param(par_to[i],Data(this_row,par_from[i]));
   }
-  prob->lsoda_init();
+  if(n > 0) prob->lsoda_init();
```
yizhang-yiz, line 192: Doesn't `any_copy` control this? This is redundant if `copy_parameters` is called only when `any_copy==true`.

kylebaron: Yeah; thanks.

yizhang-yiz (approved):

kylebaron:
@yizhang-yiz what do you think of this?  I've been benchmarking this a bit and this seems to take the edge off some of the slow down when parameters aren't changing from record to record (which I think is the main scenario where we could change things; if covariates are time-varying).

yizhang-yiz:
It makes sense. What does the benchmark say about the performance difference?

kylebaron:
# Before we changed anything


run a - `data1` is sparse input data, with constant covariate
run b - `data2` is sparse input data, without covariate
run c - `data3` is rich input with constant covariate
run d - `data4` is rich input data with no covariate

All runs have same duration and same number of output records.  There is a 
small penalty for rich input data (maybe more input processing).


```r
Unit: milliseconds
                    expr       min        lq      mean    median
 a <- mrgsim(mod, data1)  7.901581  8.800033 10.650774  9.310863
 b <- mrgsim(mod, data2)  7.968120  8.849370  9.741316  9.349273
 c <- mrgsim(mod, data3) 10.487445 12.019940 16.110642 12.752516
 d <- mrgsim(mod, data4) 10.191776 11.817085 14.856134 12.435446
```


# What we had yesterday

I made this to run slowly; simulated 3 days with every 15 minutes input record;
when there is a covariate with rich input(run c `data2`) we have to stop the 
solver every record and the time increases.  With sparse input, it doesn't make 
that big of a difference (it's like a handful of extra stops ... hard to detect).


```r
Unit: milliseconds
                    expr       min        lq      mean    median 
 a <- mrgsim(mod, data1)  7.741893  8.533279  9.791214  9.007484 
 b <- mrgsim(mod, data2)  7.795728  8.485864  9.774564  8.957434 
 c <- mrgsim(mod, data3) 35.744272 38.427414 44.985757 39.986618 
 d <- mrgsim(mod, data4) 10.219277 11.148762 13.466045 11.953646 
```



# With most recent update

With the most recent proposal; this works much faster because it doesn't restart
if the covariate value doesn't change.  If we had rich input with time-varying
covariate, you'd be back to the 44 ms run time.  But that seems to be the least
likely thing to happen.

But the penalty is that we have to do the extra work of comparing parameters 
comparison every time we update.


```r
Unit: milliseconds
                    expr       min        lq     mean    median 
 a <- mrgsim(mod, data1)  8.012418  8.991715 10.03263  9.538467 
 b <- mrgsim(mod, data2)  7.891747  8.980107 10.62298  9.505040 
 c <- mrgsim(mod, data3) 10.676537 12.227749 15.14205 12.990485 
 d <- mrgsim(mod, data4) 10.626696 11.882866 14.19894 12.798136 
```

kylebaron:
Thanks again for your help, @yizhang-yiz ; I feel good about this.

PR #763, opened 2020-11-15
Release/0.10.5

kylebaron:
https://github.com/metrumresearchgroup/mrgsolve/milestone/5


DESCRIPTION
```diff
@@ -54,7 +50,9 @@ Suggests:
     rmarkdown,
     yaml,
     knitr, 
-    data.table
+    data.table,
+    pmxTools,
```
dpastoor, line 54: where are pmxTools and spelling being used now? or were they being used but not referenced properly before. I don't see any reference to them in the code?

dpastoor (commented):
quick logistic question

DESCRIPTION
kylebaron: pmxTools in tests and I check spelling as part of my CI routine.  I check but I believe sean might have requested spelling in suggests maybe to make sure it got pulled on drone?  But I can remove it so it's not needed except where I build / check locally.

dpastoor: IMO if spelling is part of purely CI, it shouldn't be in the package dep list - in the same way that we don't add pkgdown/lintr/etc as suggested deps.

kylebaron:
I agree with you.

DESCRIPTION
kylebaron: Yep; I agree.  Just that I'm not the only one making changes to the DESCRIPTION file at this point.

kylebaron: Removed spelling from suggests but it made its way into the drone file.  I'll see if it still works.

kylebaron:
Ok; looks like it still worked.

dpastoor (approved):

Issue #775, opened 2020-12-11
Capture pp-defs by request at compile time


kylebaron:
# Summary
As a user, I want to be able to list pre-processor defined items for capture at compile time

# Tests
- inst/maintenance/unit/test-capture.R
  - capture pp directive via mread

Issue #774, opened 2020-12-11
Fix error message when `capture` variable doesn't exist


kylebaron:
# Summary
As a user, I want an informative error message when I request invalid capture items at compile time

# Tests
- inst/maintenance/unit/test-capture.R
  - capture via mread

PR #776, opened 2020-12-11
capture pp def; re-factor mread capture error messages

kylebaron:
Issues #774 and #775 

# Summary

- refactor to provide better error message when requesting capture items via `mread()`
- when using `mread_cache`, force a re-build when `capture` items are passed
  - there may be some way to cache this as well but for now you are forced to rebuild from scratch
- scrape pre-processor definitions from `$GLOBAL` and let the user `capture` those as well via `mread()`
- there is also a minor re-factor of `my_str_split` which splits a string up to `n` chunks

### DIFF
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -335,6 +335,18 @@ c_vars <- function(x,context) {
   list(vars = get_c_vars2(x,context), code = scrub_c_vars(x))
 }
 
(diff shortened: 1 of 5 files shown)


timwaterhouse (approved):
Looks good

kylebaron:
Thanks, @timwaterhouse

Issue #777, opened 2021-02-04
Handle negative times


kylebaron:
# Summary
As a user, I want mrgsolve to accept negative input times for simulations from compartmental models

# Tests
- tests/testthat/test-mrgsim.R
  - simulate non-pred with negative times is allowed
- inst/maintenance/unit/test-pred.R
  - data_set with negative times

PR #778, opened 2021-02-06
allow simulation of negative times

kylebaron:
Issue #777 
There wasn't really a good reason for not doing it before. All this does is not check.


timwaterhouse (approved):
Looks good, thanks Kyle.

Issue #783, opened 2021-02-11
Harmonize and process for initializing by program


kylebaron:
# Summary 

As a user, I want a consistent interface for programmatically initializing the following blocks: param, cmt, init, theta, omega, sigma

# Tests
- test/testthat/test-modspec.R
  - programmatic initialization
  - parse content using low-level handlers - PARAM
  - parse content using low-level handlers - INIT
  - parse content using low-level handlers - THETA
  - parse content using low-level handlers - CMT
  - parse content using low-level handlers - OMEGA / SIGMA
  - HANDLEMATRIX

PR #786, opened 2021-02-13
Refactor block handlers with emphasis on programmatic initialization

kylebaron:
Issue #783 

# Primary changes
Affected blocks: 
- PARAM
- INIT
- THETA
- SIGMA
- OMEGA

Every block has a `@object` field that allows you to name an object created in `$ENV`
Every block has an `@as_object` field that allows you to create an object locally in the code block; that is

- PARAM requires a list
- INIT requires a list
- CMT requires character
- THETA requires numeric
- OMEGA and SIGMA require matrix

- If you set `rownames` on the matrix, that becomes the labels (e.g. ECL, EV2 etc ...)

- [x] Tests

# Other refactoring
- Some changes to `[yaml]` to properly handle compartments
- A lot of linting
- Block names are returned from modelparse as is (no toupper)
- We track the exact incoming block name and track that in the parsing object
- We report using the original block name
- `evaluate_at_code` accepts the parsing environment (containing the model environment) rather than the model environment
- Fixed an issue where block numbers were off when `SET` or `ENV` were being used; they were getting zapped but no longer
- You will see changes related to `simeta`; these were already reviewed
- Removed `SET` handler; that gets handled in `mread` and we'll just handle it with the default method


tests/testthat/test-modspec.R
```diff
@@ -163,9 +149,192 @@ test_that("at options are parsed", {
...
+    "code returned the incorrect class: numeric"
+  )
+  input <- "list(a = 1, b = 2)"
+  ans <- mrgsolve:::PARAM(x = input, as_object = TRUE, env = env, pos = 3)
+  expect_is(env$param[[3]], "list")
+  expect_named(env$param[[3]])
+  
+  expect_null(env$param[[8]])  
+  env$ENV$parameters <- list(mm = 1, nn = 2)
+  ans <- mrgsolve:::PARAM(x = input, object = "parameters", env = env, pos = 8)
```
timwaterhouse, line 229: In some of these tests where you pass an object name instead of setting `as_object = TRUE`, you set `x = ""`. In other tests (like this), `x` is passed an existing object. It looks like `object` is taking precedence over `x` for these. That makes sense, but I'm just wondering what would happen if I tried using `@object` and `@as_object` in the same block. Is there a check for that before it even gets to calling these functions?

tests/testthat/test-modspec.R
```diff
@@ -163,9 +149,192 @@ test_that("at options are parsed", {
...
+
+test_that("parse content using low-level handlers - INIT", {
+  env <- mrgsolve:::parse_env(vector(mode = "list", length = 20), project = '.')
+  sup <- suppressMessages  
+  
+  input <- "c(2,2,3)"
+  expect_error(
+    sup(mrgsolve:::INIT(x = input, as_object = TRUE)), 
+    "code returned the incorrect class: numeric"
+  )
```
timwaterhouse, line 284: Should there also be a test for an unnamed list?

tests/testthat/test-modspec.R
```diff
@@ -163,9 +149,192 @@ test_that("at options are parsed", {
...
+
+test_that("parse content using low-level handlers - PARAM", {
+  env <- mrgsolve:::parse_env(vector(mode = "list", length = 20), project = '.')
+  sup <- suppressMessages  
+  
+  input <- "c(1,2,3)"
+  expect_error(
+    sup(mrgsolve:::PARAM(x = input, as_object = TRUE)), 
+    "code returned the incorrect class: numeric"
+  )
```
timwaterhouse, line 221: Should there also be a test for an unnamed list?

timwaterhouse (commented):
Really nice stuff.

tests/testthat/test-modspec.R
kylebaron: Thanks, Tim.  We'll create an error if both are supplied.

tests/testthat/test-modspec.R
kylebaron: Thanks; testing for names now.

tests/testthat/test-modspec.R
kylebaron: Thanks; testing for names now.

timwaterhouse (approved):
Looks good

PR #787, opened 2021-02-13
Store model variables in the model object

kylebaron:
# Summary
We were already looking for doubles, ints, bools, and making them global to the file. Same idea with pre-processor defines. 

This PR refactors that slightly so that we can have that information in a data structure stored in the model object. 

The information is accessed only by getting a list representation of the model (`as.list`) and looking at the `cpp_variables` item.

# Example
```r
> mod <- mcode("foo", code)
> 
> as.list(mod)$variables
     type  var  context
1     int    x preamble
2  double   CP      ode
3    bool blah    table
4 capture yick    table
```

or 

```r
> mod <- modlib("viral2")
Building viral2 ... done.
> as.list(mod)$cpp_variables
    type       var context
1 define         a  global
2 define         b  global
3 define       eps  global
4 define      epsm  global
5 double     logVT   table
6 double      logV   table
7 double     logVM   table
8 double     logIT   table
9 double logChange   table
```


timwaterhouse (commented):
You mention `variables` in the PR summary and first example. I think that's a typo and should be `cpp_variables`, but want to make sure I'm not missing something.

kylebaron:
Thanks; it was originally `variables` but I switched it to be more specific.

timwaterhouse (approved):

PR #788, opened 2021-02-13
Refactor simeta and simeps to accept n argument

kylebaron:
# Summary
Add optional integer to `simeta()` and `simeps()` 
- exception when `n` is larger than the number of eta/eps
- when `n` is positive, we resimulate everything, but only copy `eta(n)` into the vector
- when `n` is zero (default) then we resimulate everything and copy everything so that the existing behavior is retained
- I wouldn't mind requiring `n` and **only** supporting the simulation of one eta; but going with this for now to just make it possible

# Questions for reviewers
I've never used default argument before, but seems to work?  

# Waiting on
- [ ] Tests
- [ ] Review

Looking to merge this into another feature.

### DIFF
--- a/src/odeproblem.cpp
+++ b/src/odeproblem.cpp
@@ -30,10 +30,10 @@
 static Rcpp::NumericMatrix OMEGADEF(1,1);
 static arma::mat OMGADEF(1,1,arma::fill::zeros);
 
-void dosimeta(void* prob_, unsigned int n) {
+void dosimeta(void* prob_, int n) {
   odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
   arma::mat eta = prob->mv_omega(1);
-  if(n > eta.n_cols) {
+  if(n > int(eta.n_cols)) {
     throw Rcpp::exception("simeta index out of bounds", false);
   }
   if(n > 0) {
@@ -45,10 +45,10 @@ void dosimeta(void* prob_, unsigned int n) {
   }
 }
 
-void dosimeps(void* prob_, unsigned int n) {
+void dosimeps(void* prob_, int n) {
   odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
   arma::mat eps = prob->mv_sigma(1);
-  if(n > eps.n_cols) {
+  if(n > int(eps.n_cols)) {
     throw Rcpp::exception("simeps index out of bounds", false);
   }
   if(n > 0) {
--- a/src/odeproblem.cpp
+++ b/src/odeproblem.cpp
@@ -30,17 +30,31 @@
 static Rcpp::NumericMatrix OMEGADEF(1,1);
 static arma::mat OMGADEF(1,1,arma::fill::zeros);
 
-void dosimeta(void* prob_) {
+void dosimeta(void* prob_, unsigned int n) {
   odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
   arma::mat eta = prob->mv_omega(1);
+  if(n > eta.n_cols) {
+    throw Rcpp::exception("simeta index out of bounds", false);
+  }
+  if(n > 0) {
+    prob->eta(n-1,eta(0,n-1));
+    return;
+  }
   for(unsigned int i=0; i < eta.n_cols; ++i) {
     prob->eta(i,eta(0,i)); 
   }
 }
 
-void dosimeps(void* prob_) {
+void dosimeps(void* prob_, unsigned int n) {
   odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
   arma::mat eps = prob->mv_sigma(1);
+  if(n > eps.n_cols) {
+    throw Rcpp::exception("simeps index out of bounds", false);
+  }
+  if(n > 0) {
+    prob->eps(n-1,eps(0,n-1));
+    return;
+  }
   for(unsigned int i=0; i < eps.n_cols; ++i) {
     prob->eps(i,eps(0,i)); 
   }
--- a/inst/base/mrgsolv.h
+++ b/inst/base/mrgsolv.h
@@ -25,7 +25,7 @@
 #include <vector>
 #include <iostream>
 
-typedef void (*refun)(void*, unsigned int n);
+typedef void (*refun)(void*, int n);
 
 namespace mrgsolve {
 /**
--- a/inst/base/mrgsolv.h
+++ b/inst/base/mrgsolv.h
@@ -25,7 +25,7 @@
 #include <vector>
 #include <iostream>
 
-typedef void (*refun)(void*);
+typedef void (*refun)(void*, unsigned int n);
 
 namespace mrgsolve {
 /**
@@ -39,8 +39,8 @@ struct resim {
   //! resim constructor
   resim(refun x, void* y) : fun(x), prob(y){}
   resim(){}
-  void operator()() {
-    return fun(prob);
+  void operator()(int n = 0) {
+    return fun(prob, n);
   }
   
 protected: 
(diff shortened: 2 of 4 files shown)


kylebaron:
```
> library(mrgsolve)
> code <- '
+ $PARAM n = 0
+ 
+ $OMEGA 1 2 3 
+ 
+ $MAIN
+ simeta(n);
+ mrg::report("--", "resim---");
+ mrg::report("ETA 1", ETA(1));
+ mrg::report("ETA 2", ETA(2));
+ mrg::report("ETA 3", ETA(3));
+ 
+ '
> mod <- mcode("code", code, end = 3, delta = 1)
> a <- capture.output(mod %>% param(n = 0) %>% mrgsim())
-- resim---
ETA 1 1.25876
ETA 2 -1.05515
ETA 3 0.158798
-- resim---
ETA 1 0.547259
ETA 2 -0.431143
ETA 3 0.863094
-- resim---
ETA 1 -0.502001
ETA 2 0.0206066
ETA 3 -3.28179
-- resim---
ETA 1 -0.78122
ETA 2 0.841672
ETA 3 2.25358
> b <- capture.output(mod %>% param(n = 1) %>% mrgsim())
-- resim---
ETA 1 0.272231
ETA 2 0.92185
ETA 3 -0.800517
-- resim---
ETA 1 0.234305
ETA 2 0.92185
ETA 3 -0.800517
-- resim---
ETA 1 0.968213
ETA 2 0.92185
ETA 3 -0.800517
-- resim---
ETA 1 2.35072
ETA 2 0.92185
ETA 3 -0.800517
> c <- capture.output(mod %>% param(n = 2) %>% mrgsim())
-- resim---
ETA 1 -0.215285
ETA 2 -0.539847
ETA 3 -1.1875
-- resim---
ETA 1 -0.215285
ETA 2 0.747984
ETA 3 -1.1875
-- resim---
ETA 1 -0.215285
ETA 2 -1.414
ETA 3 -1.1875
-- resim---
ETA 1 -0.215285
ETA 2 -1.02739
ETA 3 -1.1875
> d <- capture.output(mod %>% param(n = 3) %>% mrgsim())
-- resim---
ETA 1 0.191843
ETA 2 -1.91717
ETA 3 -1.74346
-- resim---
ETA 1 0.191843
ETA 2 -1.91717
ETA 3 -1.30831
-- resim---
ETA 1 0.191843
ETA 2 -1.91717
ETA 3 1.46141
-- resim---
ETA 1 0.191843
ETA 2 -1.91717
ETA 3 0.983624
```

timwaterhouse (commented):
Looks reasonable.  Not sure I follow the C++ code, but resimulating multiple eta/eps values should work, right? Like
```
simeta(1);
simeta(2);
```

Also, the `cmtn` in the PR title isn't referring to a compartment number, is it?

kylebaron:
Thanks. Not sure where cmtn came from. But yeah the code you proposed should work. The test will (hopefully) confirm that.

src/odeproblem.cpp
```diff
@@ -30,17 +30,31 @@
 static Rcpp::NumericMatrix OMEGADEF(1,1);
 static arma::mat OMGADEF(1,1,arma::fill::zeros);
 
-void dosimeta(void* prob_) {
+void dosimeta(void* prob_, int n) {
```
yizhang-yiz, line 33: I feel overloading makes more sense than default args here as the two scenarios follow different logic:
```cpp
void dosimeta(void* prob_) {
  odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
  arma::mat eta = prob->mv_omega(1);
  for(unsigned int i=0; i < eta.n_cols; ++i) {
    prob->eta(i,eta(0,i)); 
  }
}
```
and
```cpp
void dosimeta(void* prob_, int n) {
  odeproblem* prob = reinterpret_cast<odeproblem*>(prob_);
  arma::mat eta = prob->mv_omega(1);
  if(n > int(eta.n_cols)) {
    throw Rcpp::exception("simeta index out of bounds", false);
  }
// is it necessary to throw when n < 0?
  if(n > 0) {
    prob->eta(n-1,eta(0,n-1));
  }
}
```
and `resim` calls the two functions as
```cpp
struct resim {
  //! resim constructor
  resim(refun x, void* y) : fun(x), prob(y){}
  resim(){}
  void operator()() {
    return fun(prob);
  }
  void operator()(int n) {
    return fun(prob, n);
  }
// ...
```

yizhang-yiz (commented):
The change is simple is enough, thought it seems overloading may be a better way.

src/odeproblem.cpp
kylebaron: Thanks, @yizhang-yiz ; I did try this approach, but I wonder if this is getting hung up on the fact that `fun` is a function pointer?

```c++
typedef void (*refun)(void*, int n);
```

So when I called `fun(prob)` in the functor,  it complained that there was only one argument. 

For this to work would I need to keep pointers to both versions and call the right one?

```c++
typedef void (*refun1)(void*);

typedef void (*refun2)(void*, int n);
```

```

Comment (yizhang-yiz) [reply src/odeproblem.cpp:33]: yes, we need to `tyedef` two function pointers. I forgot that:
```cpp
struct resim {
   refun1 fun1;
   refun2 fun2;
   void* prob;

  //! resim constructor
  resim(refun1 x, void* y) : fun1(x), prob(y){}
  resim(refun2 x, void* y) : fun2(x), prob(y){}
  resim(){}
  void operator()() {
    return fun1(prob);
  }
  void operator()(int n) {
    return fun2(prob, n);
  }
```
But this is actually where function pointer become cumbersome. Instead we can consider template
```cpp
// define two functors
struct dosimeta {
 void operator()(void* prob_, int n) {...}
 void operator()(void* prob_) {...}
};
struct dosimeps {
 void operator()(void* prob_, int n) {...}
 void operator()(void* prob_) {...}
};
```
and 
```cpp
template<typename F>
struct resim {
  //! resim constructor
  resim(F& x, void* y) : fun(x), prob(y){}
  resim(){}
  void operator()() {
    return fun(prob);
  }
  void operator()(int n) {
    return fun(prob, n);
  }
```

kylebaron:
@yizhang-yiz Could you explain the downside to the default argument approach? I'd like to keep this as simple as possible and that seems simpler than overloading.  But I admittedly don't know everything that might go wrong there.

yizhang-yiz:
The two issues I see are
- The code flow of "with n" vs "w/o n" are different. In particular, `n` has no use when it is not issue. This is why you added the `if(n>0){..return}` part to stop it early. This is not my preferred way of using default args. IMO it's for arguments such as bioavailability that every problem should have a `F` variable and one can just code it as part of the formula while allow it potentially to be 1.0. In that case, whether `F=1.0` or not doesn't affect how the code flows.
- default args are not very future-proof in the sense that if one day we decide we need another arg so that we can do `dosimeps(prob_, n, m)` the default arg approach will break down.

You can find more about pro & con of default args in google c++ style: https://google.github.io/styleguide/cppguide.html#Default_Arguments

kylebaron:
Ok;  I wouldn't do this if starting from scratch; I'm trying maintain current behavior (no argument needed) for the time being and allow people to pass in `n` as well.  Eventually that will be the only behavior and the no-argument form will be deprecated.  I'd like to take the simplest path without writing a bunch of extra code to do this.  

Is there something that will blow up if we go with the default argument?

yizhang-yiz:
I didn’t realize eventually we’re going to get rid of the first version. In that case the default arg approach completely makes sense.
  
  

  
  
>

kylebaron:
I'm sorry I didn't explain it that well.  Really, the current behavior is undesirable and not at all consistent with NONMEM does (my intention here) and it should go sooner rather than later.

yizhang-yiz (approved):

Issue #800, opened 2021-03-02
Dosing into off compartment with zero bioav does not turn it back on 


kylebaron:
# Summary 
As a user, I want to be able to turn compartments back on with doses that have zero bioavailability

# Tests
- inst/maintenance/unit/test-on-off.R
  - compartment is turned on when F is zero

kylebaron:
```r
```r

data <- c(
  ev(amt = 1, time = 0, cmt =  1),
  ev(amt = 1, time = 0, cmt =  2),
  ev(amt = 1, time = 0, cmt =  3),
  ev(amt = 0, time = 0, cmt =  1, evid = 22),
  ev(amt = 0, time = 7, cmt = -1, evid =  2),
  ev(amt = 0, time = 7, cmt = -2, evid =  2),
  ev(amt = 0, time = 7, cmt = -3, evid =  2),
  ev(amt = 1, time = 7, cmt =  3),
  ev(amt = 1, time = 7, cmt =  1),
  ev(amt = 1, time = 7, cmt =  2)
) %>% expand_observations(times = c(7,14,18))
data
```

kylebaron:
``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
code <- '
[ param ] F1 = 1, F2 = 1

[ cmt ] A B

[ main ] 
F_A = F1; 
F_B = F2;

[ ode ] 
dxdt_A = -0.1 * A;
dxdt_B =  0.1 * A;
'

mod <- mcode("test", code)
#> Building test ...
#> done.

data <- c(
  ev(amt = 1, cmt =  1, evid = 1, time = 0, F1 = 1, F2 = 1), 
  ev(amt = 0, cmt = -1, evid = 2, time = 4, F1 = 1, F2 = 1), 
  ev(amt = 0, cmt = -2, evid = 2, time = 4, F1 = 1, F2 = 1),
  ev(amt = 1, cmt =  1, evid = 1, time = 6, F1 = 1, F2 = 0),
  ev(amt = 1, cmt =  2, evid = 1, time = 6, F1 = 1, F2 = 0)
)

mrgsim(mod, data, output = "df", end = 10)
#>    ID time         A          B
#> 1   1    0 0.0000000 0.00000000
#> 2   1    0 1.0000000 0.00000000
#> 3   1    1 0.9048374 0.09516257
[...]

Issue #789, opened 2021-02-16
Call simeta for single eta


kylebaron:
# Summary
As a user, I want to be able to call simeta or simeps and re-simulate a single variate

# Tests

- inst/maintenance/unit/test-simeta.R
  - resimulate all eta
  - resimulate specific eta

Issue #782, opened 2021-02-07
Select table number in nmext


kylebaron:
# Summary
As a user, I want to be able to select a specific table number by index when using nmext

# Tests
- tests/testthat/test-nmxml.R
  - NONMEM estimates from nmext - multiple tables

kylebaron:
Please see #750

Issue #780, opened 2021-02-07
Save list of capture-able variables in model object and provide access function


kylebaron:
# Summary
As a user, I want to be able to get a listing of internal model variables

# Tests
- tests/testthat/test-modspec.R
  - inventory of internal variables

PR #804, opened 2021-03-05
Release/0.10.9

kylebaron:
# Summary 

- Dosing into a compartment that is off at the time of the dose turns the 
  compartment back on regardless of whether bioavailability is zero or non-zero 
  (this is a bug fix where the compartment was not turned on when 
  bioavailability was zero). (#800)
- `simeta()` and `simeps()` now accept an (optional) integer argument to limit
  re-simulation to single ETA or EPS values (#789)
- `as.list(mod)` output now includes a data frame of `C++` variables and 
  pre-processor definitions in the `cpp_variables` slot. (#780)
- `$PARAM`, `$THETA`, `$CMT`, `$INIT`, `$OMEGA` and `$SIGMA` blocks all include 
  the directives `@object` and `@as_object` so that block contents can be 
  specified programmatically. `@object` names an object that was coded into
  `$ENV` and `@as_object` indicates that the block contains code to realize
  the object.  See `?BLOCK_PARSE` help topic for more information on how this
  all works. (#783)
- `$NMEXT` now lets the user select between multiple tables for scraping 
  parameter estimates. (#782)


dpastoor (changes requested):
lets 🚢 !

Issue #803, opened 2021-03-04
Read NMEXT and NMXML relative to project directory


kylebaron:
# Summary
As a user, I want NMXML and NMEXT to find nonmem results relative to the project directory, not the working directory.

# Tests

- tests/testthat/test-nmxml.R
  - read nm estimates relative to cpp file

kylebaron:
Ideally this would be the default behavior; but it's not.  We'd like to eventually be at the point where both calculate the path relative to `project` but will take some time.

PR #807, opened 2021-03-13
Root directory for NMEXT and NMXML

kylebaron:
Issue #803 

There is a design doc in confluence that discusses motivation and longer term plan.

```
[ nmext ] // the default
 run = 106
 project = "../model/nonmem"
 root = "working" // implicit in the current behavior and the default new behavior

[ nmext ]
run = 106
project = "../../model/nonmem"
root = "cppfile" 

[ nmext ]
file = "../../model/nonmem/106/106.ext"
root = "cppfile"
```


timwaterhouse (commented):
Seems like a solid approach. Would absolute paths be possible as well? I'm not sure if it would add much over the "cppfile" option, but @callistosp was just introducing me to `here::here()` yesterday so I've got that on the brain. Would `here::here()` even work? Can you put any old R code into these blocks?

kylebaron:
Absolute paths should still work; that's nice part about using `setwd()` is that the absolute path works no matter what directory you are in so you shouldn't have to worry about that.  

I believe you should be able to use any old r code here; but you are introducing a dependency for the model then which I try to avoid.  But you can provide some protection on that:


```r
[ env ] 
if(!requireNamespace("here")) stop("the 'here' package is required to compile this model")

[ nmext ] 
project = file.path(here::here(), "model" , "etc"))
```

callistosp:
minor comment, `here` takes the arguments you use in `file.path`, so you can shorten that code to `here::here("model", "etc")`

PR #809, opened 2021-03-24
Save nm source in the model object

kylebaron:
When you import estimates from NONMEM, save the name of the file where the estimates were read from


timwaterhouse (approved):
Looks good

PR #821, opened 2021-05-01
stop infusions when compartment gets turned off by EVID 2

kylebaron:
# Summary

- When `cmt = -n` and `evid=2`, we turn off the compartment.  
- Current behavior: throw an error if there is an infusion into the compartment when it is turned off
- New behavior: 
  - Look for infusion off records and erase them
  - Decrement the infusion count and the total infusion rate when the infusion off record is found
  - At the end of this cycle, there will be no more infusions going into the compartment
- This will not cancel doses that were previously scheduled through `addl` 

There is a doc in confluence.


kylebaron:
@yizhang-yiz Would you be willing to review this for me?  It looks like I merged the interrupt stuff in there too; beyond the test, this is the new code: 
https://github.com/metrumresearchgroup/mrgsolve/pull/821/files#diff-38cdbb8797dc3eb63cffbd77e1502907596ae14b0764590c16c52859aeeeaad4R618-R630

yizhang-yiz (approved):
Sorry I went through it yesterday but got distracted and didn't submit.

kylebaron:
No problem; it was work in progress until just recently.  that last dplyr release messed up one of my tests so I have to get a submission for cran soon or there will be trouble. Thanks again for all your help and review.

PR #823, opened 2021-05-03
Check User Interrupt

kylebaron:
# Summary
- add `interrupt` argument to mrgsim
- defaults to 256
- we'll stop every 256th simulation record to see if the user wants to interrupt
- pass a negative integer to not check
- seems like minimal performance hit


timwaterhouse (approved):
Looks great, Kyle. Ran a few biggish test simulations and didn't see any performance hit until `interrupt` was ridiculously small.

src/odeproblem.cpp
```diff
@@ -103,6 +103,7 @@ odeproblem::odeproblem(Rcpp::NumericVector param,
   ss_flag = false;
   ssRtol = 0;
   ssAtol = 0;
+  interrupt = -4321;
```
yizhang-yiz, line 106: can we just use -1?

yizhang-yiz (approved):

yizhang-yiz (approved):

Issue #828, opened 2021-05-06
Duplicate names for capture and carry_out


kylebaron:
# Summary 
As a user, I want mrgsolve to warn me when there are duplicate names in the output and modify duplicates so that all names are unique.

# Tests
- tests/testthat/test-mrgsim.R
  - warning for duplicate output names and rename

kylebaron:
See #827

Issue #827, opened 2021-05-06
duplicate names with carry_out and capture


FelicienLL:
Hello @kylebaron ,

I noticed that when the same variable was both "captured" and "carried_out", a table with duplicate names was unexpectedly returned, which is not supported by tidyverse functions. 

It is probably not big deal, because "base" R accepts data.frames with duplicate names. And also because the problem is (maybe?) more with the user, who shouldn't use `$CAPTURE` to recover values defined in `$PARAM`, or shouldn't use both capture and carry_out for the same variable...

Anyway, do you think duplicate names could be checked when "carried out" values are copied to the simulation output, in order to avoid this behaviour?

Thank you in advance

Félicien

``` r
library(tidyverse)
library(mrgsolve)
#> 
#> Attachement du package : 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
code0 <- "
$PARAM TVCL = 1, V = 10,  BW = 70
$CMT CENT
$MAIN
double CL = TVCL * (BW / 70) ;
$PKMODEL ncmt = 1
$CAPTURE CL BW TVCL
"
mod0 <- mcode("mod0", code0)
#> Building mod0 ...
#> done.
sim <- mod0 %>%
  idata_set(expand.idata(BW = 80)) %>%
  mrgsim_df(carry_out = "BW")
head(sim)
#>   ID time BW CENT       CL BW TVCL
#> 1  1    0 80    0 1.142857 80    1
#> 2  1    1 80    0 1.142857 80    1
#> 3  1    2 80    0 1.142857 80    1
#> 4  1    3 80    0 1.142857 80    1
#> 5  1    4 80    0 1.142857 80    1
#> 6  1    5 80    0 1.142857 80    1
sim %>%
  mutate(FOO = 1)
#> Error: Can't transform a data frame with duplicate names.
```

<sup>Created on 2021-05-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

kylebaron:
Thanks; yeah it should be one or the other.

kylebaron:
@timwaterhouse any preference on what the right thing to do here?  I'd like to either warn or error rather than try to fix the names.

timwaterhouse:
[...]

PR #829, opened 2021-05-08
Documentation and handle duplicate names

kylebaron:
# Summary
- zero_re now has its own help topic
- fully removed drop_re and zero.re
- warn the user if duplicate names are found in simulated output and force the names to be unique
- tweak arguments when coercing output to data frame

See #827 and #828


timwaterhouse (approved):
Looks good

Issue #822, opened 2021-05-01
Stop infusions when compartment is turned off


kylebaron:
# Summary
As a user, I want mrgsolve to terminate all active infusions into a compartment when that compartment is turned off. 

# Tests

PR #830, opened 2021-05-10
Release/0.11.1

kylebaron:
# mrgsolve 0.11.1

- `mrgsim()` will now periodically check for user interrupt signal so that 
  long-running simulations can be stopped using `Esc` and / or `Control-C`; 
  the check interval can be modified through the `interrupt` argument to 
  `mrgsim()`, but for most applications, this shouldn't need to be changed 
  (#823)
- `mrgsim()` will issue a warning if duplicate columns are found in simulated
  output and rename duplicates using `make.names()`; thanks @FelicienLL
  for the report (#827, #828)
- Users can now turn compartments to `OFF` when they have active infusions 
  running; this was previously an error (#822)

Milestone: https://github.com/metrumresearchgroup/mrgsolve/milestone/10


dpastoor (approved):
LGTM

Issue #853, opened 2021-07-23
Recover after failed build


kylebaron:
# Summary

As a user, I want mrgsolve to return build details after build failure

# Tests
- inst/maintenance/unit/test-mread.R
  - Recover data when compile fails

PR #854, opened 2021-07-23
Feature/failure report

kylebaron:
# Summary

This `recover` argument to return information after a filed compile has been in place for a while. This refactors the code a bit. 
See #853 

you can do a basic test with the current / released version

```r
mod <- mcode("foo", "$MAIN double a = 2", recover = TRUE)
mod
str(mod)
```

The refactoring is for the most part making it easy to turn the list of lists into something that is portable (like yaml or json): turning environments into lists and fixing the class on certain items.


inst/maintenance/unit/test-mread.R
```diff
@@ -127,10 +127,26 @@ test_that("Error when code is passed as project", {
...
+    mod <- mcode("fail", code, recover = TRUE), 
+    regexp = "returning object for debugging purposes only"
+  )
+  expect_is(mod, "list")
+  expect_named(mod)
+  expect_true("mod" %in% names(mod))
+  expect_true("build" %in% names(mod))
+  expect_true("out" %in% names(mod))
+  expect_true("spec" %in% names(mod))
+  recov <- mrgsolve:::build_format_recover(data)
```
timwaterhouse, line 150: Should the `data` object here be the output of trying to compile this model? Right now this is returning `utils::data()` as character (so the test passes, but maybe this isn't doing what you're aiming for).

timwaterhouse (approved):
Looks good. Just one question about the tests.

inst/maintenance/unit/test-mread.R
kylebaron: Doh! Will fix it.

timwaterhouse (approved):

PR #857, opened 2021-07-30
Convert to data frame with Rcpp

kylebaron:
# Summary
This pr implements `as.data.frame.matrix` in Rcpp so that we can pass back a data frame from the simulation code. The changes in this PR are not user-facing. 

The implementation follows `base::as.data.frame.matrix()`.

### DIFF
--- a/src/mrgsolve.cpp
+++ b/src/mrgsolve.cpp
@@ -388,7 +388,7 @@ Rcpp::List EXPAND_OBSERVATIONS(
                             Rcpp::Named("index") = index);
 }
 
-Rcpp::List mat2df(Rcpp::NumericMatrix& x) {
+Rcpp::List mat2df(Rcpp::NumericMatrix const& x) {
   Rcpp::List ret(x.ncol());
   for(int i = 0; i < x.ncol(); ++i) {
     ret[i] = x(Rcpp::_,i);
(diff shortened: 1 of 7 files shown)


src/mrgsolve.cpp
```diff
@@ -388,6 +388,19 @@ Rcpp::List EXPAND_OBSERVATIONS(
                             Rcpp::Named("index") = index);
 }
 
+Rcpp::List mat2df(Rcpp::NumericMatrix& x) {
```
yizhang-yiz, line 391: I'm not familiar with Rcpp df API. Could you walk me through this function? Thanks. @kylebaron

kylebaron: The pattern basically follows `base::as.data.frame.matrix()`

1. Pass in a matrix
2. create a list with the same number of slots as the matrix has columns
3. Iterate through the matrix columns and copy into the list
4. `Rcpp::_` is just a way to get all the rows (or all columns)
5. The `rn(2)` bit replicates what happens in `base::.set_row_names`
6. Set the class to data frame
7. Add row names
8. Return

`as.data.frame.matrix()` does some other stuff but this does the essential steps for what I need.

yizhang-yiz: So `mat2df` doesn't take ownership of the matrix `x`'s data. Perhaps use `const` to ensure `x` is not modified? Like
```cpp
Rcpp::List mat2df(Rcpp::NumericMatrix const& x) {
```

kylebaron: Sure.

yizhang-yiz (approved):
Looks good to me!

kylebaron:
Thanks @yizhang-yiz

Issue #863, opened 2021-08-24
Using $OMEGA with NMXML


kylebaron:
# Summary
As a user, I want mrgsolve to properly deal with dropped OMEGA matrix read in by nmxml or nmet

# Tests

- tests/testthat/test-matrix.R
  - SUPERMATRIX

kylebaron:
```> code <- '
+ 
+ $OMEGA @as_object
+ ext <- read_nmext(run = 1005, project= "~/git/m4solve/inst/nonmem")
+ omega <- as.matrix(Matrix::nearPD(ext$omega)$mat)
+ omega <- unname(omega)
+ omega
+ 
+ $NMXML
+ omega = FALSE
+ project = "~/git/m4solve/inst/nonmem"
+ run = 1005
+ '
> mod <- mcode("foo", code)
 Error in h(simpleError(msg, call)) : 
  error in evaluating the argument '.x' in selecting a method for function 'omat': invalid class “omegalist” object: Length of labels (1,0) does not match the matrix rows (3). 
```

PR #865, opened 2021-08-25
fix omega collate issue

kylebaron:
# Summary

- When we use `$NMXML` to read in results, drop `OMEGA` results and ask for a different `$OMEGA` matrix in the model file, mrgsolve breaks when creating the `matlist` object.
- Modify `SUPERMATRIX` to just return the first matrix in the list when there is just one matrix.
- Add tests for `SUPERMATRIX`; these weren't already in the test suite and would have helped 
- Check types before calling `SUPERMATRIX`

# What this does

```r
> x <- list(diag(c(1,2,3)), diag(c(4,5,6,7)))
> mrgsolve:::SUPERMATRIX(x)
    [,1] [,2] [,3] [,4] [,5] [,6] [,7]
1:     1    0    0    0    0    0    0
2:     0    2    0    0    0    0    0
3:     0    0    3    0    0    0    0
4:     0    0    0    4    0    0    0
5:     0    0    0    0    5    0    0
6:     0    0    0    0    0    6    0
7:     0    0    0    0    0    0    7
```

See #863

### DIFF
--- a/R/class_matlist.R
+++ b/R/class_matlist.R
@@ -47,7 +47,7 @@ valid.matlist <- function(object) {
   }
   
   if(!x4) {
-    y <- which(sapply(object@data, det) < 0)
+    y <- which(!sapply(object@data, det) > 0)
     message("Problem with this matrix:")
     print(object@data[y])
     out <- c(out, "Invalid matrix: determinant is less than 0.")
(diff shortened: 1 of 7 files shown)


timwaterhouse (approved):
Looks good

PR #891, opened 2021-12-02
mrgsolve speaks nonmem

kylebaron:
# Summary

## General
- `[cmt]` gains `@number` option ... will make `number` compartments with a given `prefix` (like `A1`, `A2` and `A3`)
- `[ode]` gains `@audit` option ... allows users to ask that `[ode]` is audited to make sure there are ODE for every compartment
- Boolean block options can now be negated; You can write `@audit` (`TRUE`) or `@!audit` (`FALSE`)
- UTH: `gregexec` is available in `R 4.1` and would have worked but I rolled my own that more specifically did what I want; may pick up `gregexec` in the future, but I think it will be a while before we require `R 4.1`

## Related
- `mrgsolve` will now accept `THETA(n)` and use it as `THETAn`
- This is regardless of whether or not `nm-vars` is invoked
- `THETA` becomes reserved; there is an error if that name is used elsewhere in the model

## Specific

An example:

```c++
  [ plugin ] nm-vars
  [ param ] CL = 1, VC = 20, KA = 1.2, F1I = 0.8
  [ cmt ] @number 3
  [ main ] 
  F1 = F1I;
  A_0(3) = 50;
  [ ode ] 
  DADT(1) = -KA * A(1); 
  DADT(2) =  KA * A(1) - (CL/VC) * A(2);
  DADT(3) =  0;
  [ table ] 
  capture CP = A(2) / VC;

```

- Use `A(1)` not `CENT`
- Use `DADT(1)` not `dxdt_CENT`
- Use `A_0(1)` not `CENT_0`
- Use `F1` not `F_CENT`; similar for `D`, `R`, and `ALAG`
- All original variables are still available (you can still use `F_CENT` if you want)
- Use `T` in `[ode]` in place of or in addition to `SOLVERTIME`
- `A`, `A_0`, `DADT` and `T` are prohibited in `[param]` or `[cmt]` 
- `Fn`, `Rn`, `Dn`, `ALAGn` are all prohibited in `[param]`, `[cmt]` or any cpp variable
- The `[ode]` block will be audited to make sure every differential equation shows up somewhere; you can skip this by indicating `!@audit` in block options; the audit will no happen if there are no ODEs


inst/maintenance/unit/test-nm-vars.R
```diff
@@ -0,0 +1,215 @@
...
+  '
+  mod2 <- mcode("nmv-functional", code2, preclean = TRUE)
+  dose <- ev(amt = 100)
+  out1 <- mrgsim(mod1, dose, end = 48, delta = 1)
+  out2 <- mrgsim(mod2, dose, end = 48, delta = 1)
+  tol <- 0.01
+  expect_equal(out1$CP, out2$CP, tolerance = tol)
+  expect_equal(out1$GUT, out2$A1, tolerance = tol)
+  expect_equal(out1$CENT, out2$A2, tolerance = tol)
+  expect_equal(out1$RESP[1], out2$A3[1], tol = tol)
```
dpastoor, line 214:
```suggestion
  expect_equal(out1$RESP[1], out2$A3[1], tolerance = tol)
```

R/utils.R
```diff
@@ -504,9 +504,18 @@ divider_msg <- function(msg = "", width = 60) {
...
 }
 
-reg_exec_match <- function(x, pattern) {
-  m <- regexec(pattern,x)
-  regmatches(x,m)
+reg_exec_match <- function(text, pattern, ...) {
+  regmatches(text, regexec(pattern, text, ...))
+}
+
+# TODO: gregexec is now in base R as of 4.1.0
```
dpastoor, line 511: worth only conditionally creating this if R < 4.1.0?

dpastoor (approved):
tests seem pretty comprehensive

Seem ready to get out into the real world as a dev cut for people to try

R/utils.R
kylebaron: I wondered about that too. I had it all worked out with `gregexec` on 4.1 and didn't even realize it was just super new function.  My implementation does essentially what R's does (I followed the pattern), but gets to what I want more directly with a little less flexibility. So I'd rather rely on R to do the heavy lifting, but in the end the maintenance burden is about the same overall. But would make sense if I needed this in more places.

R/Aaaa.R
```diff
@@ -88,6 +88,8 @@ Reserved <- c("ID", "amt", "cmt", "ii", "ss","evid",
               paste0("pred_", c("CL", "VC", "V", "V2", "KA", "Q", "VP", "V3")),
               "double", "int", "bool", "capture")
 
+Reserved_nm <- c("A", "DADT", "A_0", "T")
```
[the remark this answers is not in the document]
kylebaron: This should only list the names that I already baked into the code. We can expand as if / when we id other variables that we want to support. This is a little over kill, but I already had this collection of `Reserved` stuff and decided to just play along.

PR #893, opened 2021-12-06
Feature/autodec

kylebaron:
# Summary

- This pr adds a plugin called `autodec`
- Searches model code, finds assignments, and declares them
  - Candidates must be valid C++ variables ... `[_a-zA-Z0-9]`
  - We also pick up symbols that contain dots and drop them in case assignments are made to, for example, `self`
- Candidate assignments are scanned for 
    - Items already existing in model definitions (e.g. parameters, compartments, etc)
    - Items declared with a type 
    - Items in reserved word list
- There is a new `modlib()` model called `nm-like` which is a basic model coded with both `nm-vars` and `autodec`
- Major re-org of the `mread()` function not a lot of change in functionality (outside of `autodec`) but stuff had to be moved   around so that some pieces deferred

### DIFF
--- a/R/modlib.R
+++ b/R/modlib.R
@@ -55,7 +55,8 @@
 ##' mod <- mread("viral2", modlib())
 ##' mod <- mread("pred1",  modlib())
 ##' mod <- mread("pbpk",   modlib())
-##' mod <- mread("1005",   modlib()) # embedded NONMEM result
+##' mod <- mread("1005",   modlib())      # embedded NONMEM result
+##' mod <- mread('nm-example", moodlib()) # model with nonmem-like syntax
 ##' 
 ##' mrgsolve:::code(mod)
 ##' }
@@ -74,7 +75,7 @@ modlib <- function(model = NULL,...,list=FALSE)  {
 modlib_models <- c(
   "pk1cmt", "pk2cmt", "pk3cmt", "pk", "pk1", "pk2", "popex",
   "irm1", "irm2", "irm3", "pred1", "emax", "tmdd", "viral1", 
-  "viral2", "effect", "1005"
+  "viral2", "effect", "1005", "nm-example"
 )
 #nocov end
 
@@ -315,4 +316,3 @@ code <- function(x) {
   }
   return(what)
 }
-
--- a/R/modlib.R
+++ b/R/modlib.R
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 - 2020  Metrum Research Group
+# Copyright (C) 2013 - 2021  Metrum Research Group
 #
 # This file is part of mrgsolve.
 #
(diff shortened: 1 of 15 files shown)


R/modlib.R
```diff
@@ -55,7 +55,8 @@
 ##' mod <- mread("viral2", modlib())
 ##' mod <- mread("pred1",  modlib())
 ##' mod <- mread("pbpk",   modlib())
-##' mod <- mread("1005",   modlib()) # embedded NONMEM result
+##' mod <- mread("1005",   modlib())      # embedded NONMEM result
+##' mod <- mread("nm-like", moodlib()) # model with nonmem-like syntax
```
timwaterhouse, line 59: typo: `moodlib()`

inst/maintenance/unit/test-misc-syntax.R
```diff
@@ -19,3 +19,17 @@ test_that("THETA(n) is allowed", {
...
   )
 })
+
+test_that("autodec + nm-vars functional test2", {
+  eps <- 1e-4
+  mod <- modlib("nm-like", delta = 0.1, preclean = TRUE, capture = "KA,INH")  
+  dose <- ev(amt = 100, D2I = 2, cmt = 2, rate = -2)
+  out <- mrgsim(mod, dose)
+  tmax <- out$time[which.max(out$A2)]
+  expect_true((tmax-2) < eps)
```
timwaterhouse, line 29: Perhaps `abs(tmax-2) < eps`?

inst/maintenance/unit/test-misc-syntax.R
```diff
@@ -19,3 +19,17 @@ test_that("THETA(n) is allowed", {
...
+  expect_true(all(out$KA==mod$THETA3))
+  expect_true("INH" %in% names(out))
+  dose <- ev(amt = 100, cmt = 1, F1I = 0.81)
+  out <- mrgsim(mod, dose)
+  expect_true((out$A1[2]-0.81*dose$amt[1]) < eps)
```
timwaterhouse, line 34: `abs(x-y) < eps`

timwaterhouse (commented):
A few minor comments, but looks good overall. I'm unreasonably excited about this one!

R/modlib.R
kylebaron: Thanks. Fixed.

inst/maintenance/unit/test-misc-syntax.R
kylebaron: Thanks.

inst/maintenance/unit/test-misc-syntax.R
kylebaron: Fixed.

R/modlib.R
kylebaron: I also pushed a change to the Makefile to get the modlib build tests "un-skipped" (it wasn't doing that); don't think this would have pick up the typo in the example, but fwiw.

kylebaron:
Comments should be addressed now; thank you for reviewing, @timwaterhouse .

timwaterhouse (approved):
Looks good

PR #901, opened 2021-12-20
Deprecate/dot ii

kylebaron:
# Summary

- There was a (mostly) undocumented argument to `ev_seq()` called `.ii` that functioned like `wait` but with origin the time of the most recent dose rather than the end of the dosing interval after the most recent dose
- The `.ii` argument has been deprecated and changed to `ii`
- Added tests for `ii` behavior
- Refactored documentation to talk more about this
- Added a bunch of safety checks ... the user must pass all event objects except for these "spacer" numerics ... if the spacer isn't named correctly we'll error
  - Check for non-event objects and error
  - Error with reminder to coerce when data.frame is found

# Example
This example illustrates the difference between `wait` and `ii`
``` r
library(mrgsolve)
#> 
#> 
#>     filter

e1 <- ev(amt = 100, ii = 12, addl = 24)
e2 <- ev(amt = 50, ii = 24, addl = 100)

seq(e1, wait = 100, e2)
#> Events:
#> 1    0 100 12   24   1    1
#> 2  400  50 24  100   1    1
seq(e1, ii = 100,   e2)
#> Events:
#> 1    0 100 12   24   1    1
#> 2  388  50 24  100   1    1
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


timwaterhouse (approved):
lgtm

PR #907, opened 2022-01-04
Allow autodec skip; fix audit problem

kylebaron:
# Summary 

- Implements an `autodec` escape hatch, allowing users to list variables to NOT declare in case they are getting picked up
  - This is set in `$ENV MRGSOLVE_AUTODEC_SKIP = c(...)`
  - Some documentation was added to `autodec_clean()` since that was getting more complicated; this is not exported and the docs are for people reading the code only

- Also: I'm just seeing that there is an auditing problem that was introduced with `nm-vars` plugin; basically, we need to skip auditing ODEs for standard models when the `nm-vars` plugin is in play
  - Added `mread.env` slots (logical) indicating whether or not `nm-vars` are in play
  - This is checked prior to standard auditing
  - I plan to harmonize this a bit more down the road ... less piecemeal

- Added a check for reserved names when `nm-vars` in invoked along with `autodec`; this wasn't in place before
  - For example, you should be able to use `A` when `nm-vars` isn't in play; but when it is, it should be reserved and error when `A` is declared
  - This most recent check picks up `A` when it is found by `autodec`

### DIFF
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -631,6 +631,8 @@ parse_env <- function(spec, incoming_names = names(spec),project,ENV=new.env()) 
   mread.env$annot <- vector("list", n)
   mread.env$ode   <- vector("list", n)
   mread.env$audit_dadt <- FALSE
+  mread.env$using_autodec <- FALSE
+  mread.env$`using_nm-vars` <- FALSE
   mread.env$namespace <- vector("list", n)
   mread.env$capture <- vector("list", n)
   mread.env$error <- character(0)
@@ -764,6 +766,13 @@ autodec_vars <- function(code, blocks = NULL) {
 
 #' Clean up `autodec` candidates
 #' 
+#' @param vars candidates
+#' @param rdefs compartments and parameters that will be implemented with 
+#' C++ pre-processor defines
+#' @param build the model `build` object; this contains variables that will 
+#' be globally declared
+#' @param skip additional names to scrub
+#' 
 #' @details 
 #' 
 #' Remove
@@ -775,11 +784,12 @@ autodec_vars <- function(code, blocks = NULL) {
 #' @md
 #' @keywords internal
 #' @noRd
-autodec_clean <- function(vars, rdefs, build) {
+autodec_clean <- function(vars, rdefs, build, skip = NULL) {
   rdefs <- strsplit(rdefs, " ", fixed = TRUE)
   rdefs <- s_pick(rdefs, 2)
   cpp <- build[["cpp_variables"]][["var"]]
-  vars <- setdiff(vars, c(Reserved, rdefs, Reserved_nm, cpp))  
+  vars <- setdiff(vars, c(Reserved, rdefs, Reserved_nm, cpp))
+  vars <- setdiff(vars, skip)
   vars
 }
 
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -631,8 +631,9 @@ parse_env <- function(spec, incoming_names = names(spec),project,ENV=new.env()) 
   mread.env$annot <- vector("list", n)
   mread.env$ode   <- vector("list", n)
   mread.env$audit_dadt <- FALSE
-  mread.env$using_autodec <- FALSE
   mread.env$`using_nm-vars` <- FALSE
+  # TODO:
+  #mread.env$using_autodec <- FALSE
   mread.env$namespace <- vector("list", n)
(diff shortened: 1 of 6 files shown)


R/modspec.R
```diff
@@ -631,6 +631,9 @@ parse_env <- function(spec, incoming_names = names(spec),project,ENV=new.env())
   mread.env$annot <- vector("list", n)
   mread.env$ode   <- vector("list", n)
   mread.env$audit_dadt <- FALSE
+  mread.env$`using_nm-vars` <- FALSE
+  # TODO:
+  #mread.env$using_autodec <- FALSE
```
kylebaron, line 636: Hold off on setting this for now.

R/modspec.R
```diff
@@ -764,6 +767,13 @@ autodec_vars <- function(code, blocks = NULL) {
...
 #' Clean up `autodec` candidates
 #' 
+#' @param vars candidates
+#' @param rdefs compartments and parameters that will be implemented with 
+#' C++ pre-processor defines
+#' @param build the model `build` object; this contains variables that will 
+#' be globally declared
+#' @param skip additional names to scrub
+#' 
 #' @details
```
kylebaron, line 777: internal docs

R/mread.R
```diff
@@ -369,15 +369,22 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
...
+    auto_skip <- cvec_cs(ENV[["MRGSOLVE_AUTODEC_SKIP"]])
     autov <- autodec_vars(spec, blocks = auto_blocks)
-    autov <- autodec_clean(autov, rdefs = rd, build = build)
+    autov <- autodec_clean(
+      autov, 
+      rdefs = rd, 
+      build = build, 
+      skip = auto_skip
+    )
+    autodec_nm_vars(autov, mread.env)
```
kylebaron, line 383: This is the new check that looks for variables that we're trying to autodec but are reserved when `nm-vars` is invoked.

R/mread.R
```diff
@@ -423,11 +430,14 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
...
   
   # This must come after audit
+  # TODO: harmonize with other audit process
   if(!has_name("ODE", spec)) {
     spec[["ODE"]] <- "DXDTZERO();"
-  } else if(mread.env[["audit_dadt"]]) {
-    spec[["ODE"]] <- c(spec[["ODE"]], ode)
-    audit_spec(x, spec, warn = warn)
+  } else {
+    if(mread.env[["audit_dadt"]] && !mread.env[["using_nm-vars"]]) {
```
kylebaron, line 437: This fixes the audit issue that we found later on; I have on TODO list to harmonize with other audits.

R/nm-mode.R
```diff
@@ -118,6 +118,22 @@ audit_nm_vars <- function(x, param, init, build, nmv, env) {
   return(invisible(TRUE))
 }
 
+autodec_nm_vars <- function(x, env) {
```
kylebaron, line 121: This check picks up what was missed when handling `nm-vars`.

tests/testthat/test-modspec.R
```diff
@@ -521,3 +521,19 @@ test_that("autodec models with nm-vars", {
   expect_equal(cpp$var, c("km","err", "cl", "v2", "ka", "CP"))
   expect_equal(cpp$context, c("main", "table",  rep("auto", 4)))
 })
+
+test_that("autodec variables can be skipped", {
```
kylebaron, line 525: Undocumented.

R/mread.R
```diff
@@ -369,15 +369,22 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
       env = mread.env
     )
   }
-  
   # autodec
   if("autodec" %in% names(plugin)) {
     auto_blocks <- c("PREAMBLE", "MAIN", "PRED", "ODE", "TABLE")
+    auto_skip <- cvec_cs(ENV[["MRGSOLVE_AUTODEC_SKIP"]])
```
kylebaron, line 375: Check if there are any variables to whitelist from `autodec`

timwaterhouse (approved):
This looks good to me. I don't have any big ideas about how to handle this any other way for now.

kylebaron:
Thanks, @timwaterhouse; let's give this a try in the wild and see what happens. Even if something needs to be adjusted down the road, I'm pretty happy with it ... didn't think it would get this far.

PR #909, opened 2022-01-05
Warn when simetan or simepsn is requested ETA or EPS are correlated

kylebaron:
# Summary

- look for `simeta(n)` in `$MAIN` or `simeps(n)` in `$TABLE`
- warn if any off diagonals for `OMEGA` or `SIGMA` are non zero


timwaterhouse (approved):
Looks good

kylebaron:
Thanks @timwaterhouse ; just noting here that users won't get warned in case they code $OMEGA as a diagonal matrix but then later update with non-zero off diagonals. I think this still covers most of the usage and still a good first step.

PR #919, opened 2022-01-17
Feature - events with data-like names

kylebaron:
# Summary
- `ev` objects render with lower case names, including `time` in the simulated output
- This implements a new flag in the `ev` object that will render nmtran-like names in upper case, like we usually do in DATA
- The arguments for the constructor and the names in the underlying object are still lower case
- Any manipulations to the object also use lower case names
- Names are turned to uppercase when `as.data.frame()` is called or when the object is used for simulation
- There's a quirk when using `carry_out` for nmtran names; these always come out as you name them; this is quite a special case and there is involvement in how the C++ code is handled; this isn't new and the behavior will remain going forward until there's a appetite to change it
- All event object operations remain the same, but when `data.frames` are returned, names will be rendered accordingly

This PR also revisits `lctran()` which converts upper case nmtran names to lower case. We add `uctran()` to make the names upper case instead. Also these functions now warn when there are both upper and lower case nmtran names. This would be a complimentary workflow to `evd()`. 

# Example
``` r
library(mrgsolve)
#> 
#> 
#>     filter
e <- evd(amt = 100, ii = 24, addl = 5)
e
#> Events Data:
#> 1    0 100 24    5   1    1
as.data.frame(e)
#>   TIME AMT II ADDL CMT EVID
#> 1    0 100 24    5   1    1
mrgsim(house(), e)
#> Model:  housemodel 
#> Dim:    482 x 7 
#> Time:   0 to 120 
#> ID:     1 
#>     ID TIME    GUT  CENT  RESP    DV    CP
#> 1:   1 0.00   0.00  0.00 50.00 0.000 0.000
#> 2:   1 0.00 100.00  0.00 50.00 0.000 0.000
#> 3:   1 0.25  74.08 25.75 48.68 1.287 1.287
#> 4:   1 0.50  54.88 44.50 46.18 2.225 2.225
#> 5:   1 0.75  40.66 58.08 43.61 2.904 2.904
#> 6:   1 1.00  30.12 67.83 41.38 3.391 3.391
#> 7:   1 1.25  22.31 74.74 39.58 3.737 3.737
#> 8:   1 1.50  16.53 79.56 38.18 3.978 3.978
```

<sup>Created on 2022-01-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


timwaterhouse (approved):
Looks good. It's a shame about the `carry_out` behaviour, but I don't see it being a huge problem.

kylebaron:
Thanks, @timwaterhouse ; I think the carry out behavior should be fixable; it works like that partly due to lazy implementation / workaround a long time ago when we started accepting either upper or lower case names.

PR #930, opened 2022-02-21
Refactor to use mod directly when passing into the simulation

kylebaron:
# Summary

The theme if this PR is to pass `mod` into `DEVTRAN` and extract information from the model object to use in the C++ simulation layer. 

Data are extracted from an `Rcpp::S4` object using `.slot()`. Care must be taken because we are working on pointers when we access through `.slot()`; the simulation code can change the values in `mod`. We take care to `Rcpp::clone()` where needed. 

# Specifics
- `MAKEMATRIX` now takes a matlist object 
- There was a biggish list called `parin` which got passed in to the C++ code; this was basically a convenience shuttle so we didn't have to change the signature all the time; most (not all) of this is no longer needed since we were copying a bunch of items from the model object, into that list and then using that data from there
- You will see `mod` passed to a lot of functions because we are extracting info from that `S4` object rather than the `parin` list
- When checking if the model is loaded, let's just check `main` and take that; I don't think there is really a need to check that all the functions are loaded
- `copy_parin` now takes both the `parin` (we still need that) and the model object
- We had C++ functions called `omega()` and `sigma()` which registered those matrices; now, we pass the model object and call `MAKEMATRIX` on OMEGA and SIGMA there
- One super important detail: when we access items in `mod`, we call this `.slot()` method; when you do this, you are dealing with a pointer and `mod` items can get modified in place; we have to be careful to clone stuff that we want to modify
- We also had this really dumb system for passing capture indices plus the total number of captures; it was one vector and the first position was the total number and the remaining positions were the ones to actually take; the PR undoes that 
- Not really any need to look at RcppExports.cpp


.Rbuildignore
```diff
@@ -1,5 +1,5 @@
 ^run-*\.R$
-^test-*\.R$
+^./test-.*\.R$
```
kyleam, line 2: I wasn't able to figure out the point of this change.  Doesn't the `./` anchor this on a single character followed by a slash?  Intentional?

R/RcppExports.R
```diff
@@ -1,8 +1,8 @@
 # Generated by using Rcpp::compileAttributes() -> do not edit by hand
 # Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
 
-DEVTRAN <- function(parin, inpar, parnames, init, cmtnames, capture, funs, data, idata, OMEGA, SIGMA, envir) {
-    .Call(`_mrgsolve_DEVTRAN`, parin, inpar, parnames, init, cmtnames, capture, funs, data, idata, OMEGA, SIGMA, envir)
+DEVTRAN <- function(parin, funs, data, idata, mod) {
```
kyleam, line 4: This is a very satisfying signature reduction.  Same for `TOUCH_FUNS` below.

R/class_matlist.R
```diff
@@ -47,14 +51,14 @@ valid.matlist <- function(object) {
   }
   
   if(!x4) {
-    y <- which(!sapply(object@data, det) > 0)
+    y <- which(!vapply(object@data, det, 1.23) > 0)
```
kyleam, line 54: Why 1.23?

kyleam (commented):
I'm lacking the combination of mrgsolve and Rcpp chops to be that helpful of a reviewer, though both the "Summary" and "Specifics" part of PR description were very helpful for getting a sense for what was going on in this PR.

Read through this a few times.  I can't say I really grok everything, but it looks like a nice direction to go.

R/class_matlist.R
```diff
@@ -16,22 +16,26 @@
 # along with mrgsolve.  If not, see <http://www.gnu.org/licenses/>.
 
 # this won't get included in coverage since it's copied into the class def
+
+
+checkSymmetric <- function(h) isSymmetric(unname(h))
```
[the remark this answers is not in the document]
kylebaron: Thanks; I removed that and will circle back on this symmetric issue. IMO it can be handled once on load or if the matrices are ever updated.

R/class_matlist.R
kylebaron: Just to indicate that we're expecting a single numeric value.  I updated the call to name the arguments.

kylebaron:
@kyleam Thanks for looking at this.  It's a lot of under the hood so in a way covered by a lot of the existing tests and I tried to add tests to cover behavior that we might not be expecting.

I intentionally left this out of the last PR so we could try to merge this and just use it in development for a while to see what else might pop out. 

Kyle

PR #936, opened 2022-03-07
add nm-math to nm-vars include; modify functional test

kylebaron:
Only including `SQRT`, `EXP` and `LOG` for now.

Issue #935, opened 2022-03-07
Data like event objects


kylebaron:
# Summary 
As a user, I want mrgsolve to render names in event objects in a data-like manner - with all caps for NMTRAN data items

# Tests
- tests/testthat/test-evd.R
  - create evd object
  - evd object has all lower case names
  - evd object realize names
  - evd object simulated names
  - evd object carry out tran names
  - coerce ev object to evd
  - ev operations with evd objects
  - Convert names to lower case with lctran
  - Convert names to upper case with uctran

Issue #931, opened 2022-02-28
NONMEM math


kylebaron:
# Summary
As a user, I want mrgsolve to recognize `EXP`, `LOG`, `SQRT` when writing models with `nm-vars`

# Tests
- inst/maintenance/unit/test-nm-vars.R
  - nm-vars functional test

Issue #927, opened 2022-02-01
self.stop_id() should fill NA; self.stop_id_cf() should carry forward


kylebaron:
# Summary
As a user, I want mrgsolve to stop the simulation for a given id and fill with either NA or the last value, as directed.

# Tests
- inst/maintenance/unit/test-stop-id.R
  - Stop the current ID and carry forward
  - Stop the current ID and fill NA
  - Stop the entire simulation

PR #939, opened 2022-03-12
Release/1.0.1

kylebaron:
- [ ] release version number in NEWS
- [ ] release version number in DESCRIPTION

# Summary 
- Add `LOG()`, `EXP()`, `SQRT()` macros when `nm-vars` plugin is invoked 
  #931, #936.

- Use `evd()` to create an event object which renders nmtran names 
  in upper case (e.g. `TIME` rather than `time`) #935, #919.

- Fixed bug where `rate` was not getting set for modeled events #934.

- Fixed bug where `self.stop_id()` and `self.stop_id_cf()` had reversed
  behavior as documented #927, #928.

- Refactored EVID=3 behavior to leave `NEWIND` as-is #934.


.drone.yml
```diff
@@ -1,54 +1,209 @@
...
 
 steps:
-- name: Pull mpn container from ECR
+- name: pull
   image: omerxx/drone-ecr-auth
+  commands:
+  - $(aws ecr get-login --no-include-email --region us-east-1)
+  - docker pull 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.1:latest
+  - docker pull 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.0:latest
+  - docker pull 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-3.6:latest
```
seth127, line 17: You can delete this line because you don't actually use this container.

As discussed offline, the `r-dev-ci-mpn-3.6:latest` container hasn't been built since May 2021 so it doesn't have the required versions of deps. I think it's fine to drop it from here completely, especially since you're testing 3.6 against CRAN-latest below.

.drone.yml
```diff
@@ -1,54 +1,209 @@
...
+  - make drone
+
+- name: "Check package: R 4.0"
+  pull: never
+  image: 906087756158.dkr.ecr.us-east-1.amazonaws.com/r-dev-ci-mpn-4.0:latest
+  commands:
+  - R -s -e 'devtools::install_deps(upgrade = '"'"'never'"'"', dependencies = TRUE)'
+  - make drone
+
+# - name: "Check package: R 3.6"
```
seth127, line 36: You can also delete these commented lines.

seth127 (approved):
All looks good. A few comments you can take or leave for now about drone cleanup.

PR #940, opened 2022-03-16
re-config .drone file

kylebaron:


kyleam (approved):
Changes look fine to me.  As discussed offline with you and @seth127, I think restricting to `:cran-latest` is okay. (If we get burned by that, we can of course revisit.)

PR #955, opened 2022-03-26
Fix/as data set

kylebaron:
# Summary
- `as_data_set()` didn't properly convert names when only one object was passed in
- Adds `expand.evd` and `evd_expand()` to create uppercase data sets
- Refactor `uctran()` and `lctran()` to be generic, receiving either data.frame or event object
  - The event object method doesn't _actually_ changes names; but rather changes the value of `case` so that when the names are rendered they are either upper or lower case
  - This required tweaking of the `as.ev.ev` method
- Removing some unneeded tidyverse stuff (no reason to use it)


timwaterhouse (approved):
Looks good!

R/data_set.R
```diff
@@ -167,12 +167,13 @@ setMethod("data_set", c("mrgmod", "missing"), function(x, object, ...) {
 #' If both lower and upper case versions of the name are present in the data 
 #' frame, no changes will be made.
```
shenc-metrum, line 168: Just a question, Kyle. Are the columns listed above the only data items will be controlled by `lctran()` or `uctran()`. If so, wouldn't be better to also include things like `ID`, `MDV`? 
Additionally, any specific reason for this: "If both lower and upper case versions of the name are present in the data frame, no changes will be made"

shenc-metrum (commented):
Looks good, Kyle!
I just left a question in `data_set.R`

R/data_set.R
kylebaron: Thanks, Shen. 

1. There is no MDV
2. ID is always capital (that was one of the motivations for this)
3. We can't have both `amt` and `AMT` in the data set and then turn `amt` into `AMT`; this really shouldn't happen and we wait until the data is submitted for simulation where this will turn into an error.

shenc-metrum: Makes sense. Thanks Kyle!

PR #968, opened 2022-05-07
change name of template parameters to work with nm-vars

kylebaron:
# Summary 
When we added `T` macro for NONMEM syntax, it interfered with how the templates were written when both plugins were invoked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,9 @@ test1:
 test2:
 	Rscript -e 'testthat::test_dir("inst/maintenance/unit")'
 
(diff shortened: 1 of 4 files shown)


yizhang-yiz:
@kylebaron a coupled of Qs.

How do templates interfere in this case exactly? 

Other than in the example `mrgx::get<Rcpp::NumericMatrix>("mymatrix", self)`, what other types can `_T__` assume? It seems the function body doesn't use `_T__` at all. In that case can we use `auto` for return?

PR #969, opened 2022-05-07
Only process lag time if EVID is 1 or 4

kylebaron:
# Summary 

- ALAG was affecting timing of any event record
- The PR limits this so that only EVID 1 or EVID 4 records are offset by ALAG

Test in maintenance test-z-alag-f.R

Issue #964, opened 2022-05-06
ALAG should only apply to EVID 1 or 4 


kylebaron:
Looks like it is getting applied to EVID 3

The behavior looks weird because the compartment is reset at the event time and then resets again for the lagged event.

Issue #956, opened 2022-03-28
Deal with negative eigenvalue in model matrix


kylebaron:
# Summary
As a user, I want mrgsolve to handle small negative eigenvalues when simulating from model matrices

# Tests
- tests/testthat/test-rng.R
  - test-rng mrgsolve simulated with negative eigenvalue

PR #974, opened 2022-05-12
Release/1.0.4

kylebaron:
- Fix bug where `as_data_frame()` was not properly working when leading event
  object was `evd` type (#948, #955).

- Add `uctran()` to convert nmtran data names to upper case (#955). 

- Both `lctran()` and `uctran()` are generic and work on `data.frame` or 
  event (`ev`) objects (#949, #955). 

- Fix bug where data records with `EVID = 3` were getting shifted in time by
  `ALAG` (#964, #969).

- Small negative eigenvalues from `OMEGA` or `SIGMA` are set to zero in 
  multivariate normal simulation of `ETA` and `EPS`, following the pattern
  seen in `MASS::mvrnorm()` (#956, #957).

- Fixed bug where template parameters (`T`) were getting modified when `nm-vars`
  plugin was used in conjunction with `mrgx` plugin (#965, #968).

# Tests

Starting to collect user stories and tests in a yaml file; going to start building this up but it does not fit into the mainstream testing / stories setup at this point.


inst/stories.yaml
```diff
@@ -0,0 +1,35 @@
+# Please add stories at the top ------------------------------------------
+
+SLV-S004: 
+  name: evid 3 with lag time
+  description: > 
+    As a user, I want mrgsolve to run evid=3 reset at the given time, regardless
+    of teh value of ALAG.
```
kyleam, line 7:
```suggestion
    of the value of ALAG.
```

kyleam (approved):
Went over the release-specific bits, and they look good to me.  Left one comment on a stories.yaml typo.

inst/stories.yaml
kylebaron: Thanks @kyleam

PR #978, opened 2022-05-14
refactor parameter list updates 

kylebaron:
# Summary

## parameter update
- There was some very old code to update the model parameter list (via `param()`)
- It was really inefficient and called itself a bunch of times
- The update took way too long
- This simplifies things and adds some additional tests and constraints 
- Part of the inefficiency was around calling `methods::validObject()` on the new parameter object on every update; this is more needed on construction and with the modern setup, parameter list construction only happens when the model object is constructed (used to not be that way); so rather than calling `validObject()` let's just to a basic check on the **incoming** object providing names / values to drive the update

# Testing

- most of this was tested before
- I refactored some of the tests and added some
- Created a single user story for updating parameters in the model and listed these tests


kylebaron:
@kyleam I'll have to re-number some tests from the other PR. Not sure how you've been doing this but thinking it'll be easier to merge and then untangle it. But if you have ideas, please let me know.

R/param.R
```diff
@@ -18,121 +18,159 @@
...
+#' For both uses, the return value is a `parameter_list` object. For the 
+#' former use, `param` is usually called to print the parameters to the 
+#' screen, but the `parameter_list` object can also be coerced to a list 
+#' or numeric R object.
+#' 
+#' Use `allparam()` to get a `parameter_list` object including 
+#' both model parameters and data items listed in `$FIXED`.
+#' 
+#' The update to parameters can be permissive (candidates with names that don't
+#' exist in the parameter list are sliently ignored) or strict (all candndates
```
kyleam, line 53: typo: candndates -> candidates

R/utils.R
```diff
@@ -77,11 +77,20 @@ combine_list <- function(left, right) {
...
 }
 
-update_list <- function(left, right) {
+update_list <- function(left, right, context = NULL) {
   if(!all(is.list(left),is.list(right))) {
-    stop("input are not lists")
+    msg <- "input are not lists"
+    context <- paste0("[", context, "]")
+    msg <- paste(context, msg)
+    stop("msg")
```
kyleam, line 85: code typo: "msg" -> msg

inst/stories.yaml
```diff
@@ -1,4 +1,14 @@
 # Please add stories at the top ------------------------------------------
+SLV-S005:
```
kyleam, line 2: In the merge conflict at 7513cf7c (Merge branch 'develop' into refactor/param, 2022-08-16), this should have been bumped to SLV-S006.  And this will have yet another conflict with PR #1008.  Unfortunately, in-tree stories very much breed conflicts :/

kyleam:
> I'll have to re-number some tests from the other PR. Not sure how you've been doing this but thinking it'll be easier to merge and then untangle it. But if you have ideas, please let me know.

Yeah, assuming the other PR is merged first and it doesn't lead to actual text conflicts, merging this and then untangling works.  If there are actual text conflicts, you'll need to bring `develop` into this PR to resolve.

kyleam:
Rationale makes sense to me, and didn't spot any issues aside from the minor ones I mentioned inline.

R/param.R
kylebaron: Thanks; fixed.

R/utils.R
kylebaron: ugh ... fixed.

R/param.R
```diff
@@ -18,121 +18,159 @@
...
+    wstop("[param-update] invalid object to update parameter list.")
+  }
+  
+  if(is.data.frame(.y) && nrow(.y) != 1) {
+    stop(
+      "[param-update] data.frame input should have exactly one row.", 
+      call.=FALSE
+    )
+  } else {
+    .y <- as.list(.y)
```
kylebaron, line 127: @kyleam I think this is where it's crumping.

PR #987, opened 2022-05-20
Dynamic capture of nm-var items

kylebaron:
# Summary

When checking for valid (dynamic) capture names, we weren't including variables from `nm-vars` as valid items to capture.  For example, `F1` is a special, `nm-vars` variable and if I tried to capture this dynamically (on model load) it would not recognize it as a valid capture item. 

This pull request adds those `nm-vars` variables into the valid capture list.

We also create a user story for dynamic capture; number and collect existing and new tests under that story.


kyleam (approved):
LGTM

PR #1008, opened 2022-08-15
Drop/long int

kylebaron:
# Summary

When `data.table` (and maybe `readr`) read in `.csv` files, they can create columns with class `integer64`. When the data frame is converted to matrix to pass into C++ layer, it can't handle that data ... pretty much undefined behavior (I'm not sure what all is going on but it looks like undefined).  The fix for this is to coerce with `as.double`.

I am handing this by dropping any column that has a class attribute set. So valid data are bare numerics.

I am also refactoring the message that appears when columns are dropped (like `character`) so that the user is told the name of the column that is getting dropped along with the class of that column. 

Test is added specifically for `integer64` data. Previous tests for dropping `character` columns have been updated to look for the new message.


kylebaron:
# Broken

```r
> data <- fread("ID,KA\n1,100000020200")
> data$TIME <- 1
> data$CMT <- 1
> mod <- house()
> mrgsim(mod, events = data)
Error: event record cmt must be between 1 and 3: 
 ID: 4.94066e-324, row: 1, cmt: 0, evid: 4.94066e-324
```

# Fixed

```r
> data <- fread("ID,KA\n1,100000020200")
> data$TIME <- 1
> data$CMT <- 1
> 
> mod <- house()
> mrgsim(mod, events = data)
[data-set] dropped column: KA (integer64)
Model:  housemodel 
Dim:    482 x 7 
Time:   0 to 120 
ID:     1 
    ID time GUT CENT RESP DV CP
1:   1 0.00   0    0   50  0  0
2:   1 0.25   0    0   50  0  0
3:   1 0.50   0    0   50  0  0
4:   1 0.75   0    0   50  0  0
5:   1 1.00   0    0   50  0  0
6:   1 1.00   0    0   50  0  0
7:   1 1.25   0    0   50  0  0
8:   1 1.50   0    0   50  0  0
```

R/mrgindata.R
```diff
@@ -81,6 +81,18 @@ convert_character_cmt <- function(data, mod) {
   return(data)
 }
 
+signal_drop <- function(dm, x, to_signal, context) {
+  drop <- setdiff(names(x), dimnames(dm)[[2]])
+  drop <- intersect(drop, to_signal)
+  for(d in drop) {
+    type <- paste0(class(x[[d]]), collapse = ",")
+    msg <- c(context, " dropped column: ", d, " (", type, ")")
+    msg <- paste(msg)
```
kyleam, line 90: Hmm, what's this `paste` doing?  Won't the result be the same as the input?


```
> x <- c("foo", "bar")
> identical(x, paste(x))
[1] TRUE
```

kyleam (approved):
Looks good.  (Left a comment about a `paste` call, but even if I'm right, it doesn't change the behavior in any way.)

R/mrgindata.R
kylebaron: I see ... I think I was intending to collapse that vector but message is doing it for me. Thanks.

PR #1025, opened 2022-10-30
Let NMEXT and NMXML assume shared control stream root for NONMEM and mrgsolve

kylebaron:


R/utils.R
```diff
@@ -645,3 +645,27 @@ mod_first <- function(cl) {
   msg <- sprintf("the first argument to %s must be a model object",fun)
   wstop(msg)
 }
+
+#' Parse a file name
+#' @param x a file name to parse.
+#' @return 
+#' A list with `ext`, `root` and `has_ext`.
+#' @details
+#' This follows the approach used in [tools::file_path_sans_ext].
```
kyleam, line 654: Probably fine in this context, but `tools::file_path_sans_ext` is a bit stricter regarding file names starting with a dot:

``` r
tools::file_path_sans_ext(".foo")
#> [1] ".foo"
```

It looks like only the `root` value is used in this PR, so I'm curious why it's worth adding a custom function instead of using `tools::file_path_sans_ext` (but the code looks fine to me, aside from perhaps the ".foo" behavior).

tests/testthat/test-nmxml.R
```diff
@@ -353,3 +353,19 @@ test_that("nm source file is available via as.list", {
...
+  expect_is(mod, "mrgmod")
+  nmext_file <- basename(as.list(mod)[["nm_import"]])
+  expect_equal(nmext_file, "1005.ext")
+})
+
+test_that("use cpp file stem as nm run number nmxml [SLV-TEST-0022]", {
+  skip_if_not(file.exists("nm/cppstem-nmxml/1005.cpp"))
+  mod <- mread("1005", project = "nm/cppstem-nmxml")
+  expect_is(mod, "mrgmod")
+  nmext_file <- basename(as.list(mod)[["nm_import"]])
```
kyleam, line 369: Doesn't matter for the correctness of the test, but this should be `nmxl_file`, right?

R/modspec.R
```diff
@@ -631,9 +631,10 @@ scrape_and_call <- function(x,env,pass,...) {
 }
 
 dump_opts <- function(x,env,block,...) {
-  hasopt <- unique(c(grep(">>", x, fixed=TRUE),grep("@", x, fixed=TRUE))) 
+  hasopt <- unique(c(grep(">>", x, fixed = TRUE), grep("@", x, fixed = TRUE))) 
+  if(length(hasopt)==0) return(x)
+  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl = TRUE)
   if(length(hasopt)==0) return(x)
-  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl=TRUE)
   x[-hasopt]
```
kyleam, line 638: The PR description mentions:

> dump_opts had to be refactored to allow an @ sign in quotes passed as run

... but, looking at these changes, it's not popping at to me what change accomplished that.

And looking over this function, I'm getting turned around. I *think* it's supposed to be removing "\>\> ..." and "@ ..." values, but I don't see how it does that. It looks wrong to me that it calculates the `hasopt` indices from the *subsetted* `x` but then uses `-hasopt` to index the full `x`.

Here's what I was playing around (moving the unused arguments):

<details>
<summary>snippet</summary>

``` r
dump_opts_old <- function(x) {
  hasopt <- unique(c(grep(">>", x, fixed=TRUE),grep("@", x, fixed=TRUE)))
  if(length(hasopt)==0) return(x)
  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl=TRUE)
  x[-hasopt]
}

dump_opts <- function(x) {
  hasopt <- unique(c(grep(">>", x, fixed = TRUE), grep("@", x, fixed = TRUE)))
  if(length(hasopt)==0) return(x)
  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl = TRUE)
  if(length(hasopt)==0) return(x)
  x[-hasopt]
}

dump_opts_old(c("a", "b"))
#> [1] "a" "b"
dump_opts(c("a", "b"))
#> [1] "a" "b"

dump_opts_old(c("a", "@b"))
#> [1] "@b"
dump_opts(c("a", "@b"))
#> [1] "@b"

dump_opts_old(c("a", "@b", ">>c"))
#> [1] ">>c"
dump_opts(c("a", "@b", ">>c"))
#> [1] ">>c"

dump_opts_old(c("a", "@b", "c"))
#> [1] "@b" "c"
dump_opts(c("a", "@b", "c"))
#> [1] "@b" "c"
```
</details>

tests/testthat/test-nmxml.R
kylebaron: Yes; I fixed it for clarity.

R/modspec.R
kylebaron: Here's reprex

``` r
code <- c('root = "@cppstem"', 'run = 1002')

dump_opts_old <- function(x) {
  hasopt <- unique(c(grep(">>", x, fixed=TRUE),grep("@", x, fixed=TRUE)))
  if(length(hasopt)==0) return(x)
  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl=TRUE)
  x[-hasopt]
}

dump_opts <- function(x) {
  hasopt <- unique(c(grep(">>", x, fixed = TRUE), grep("@", x, fixed = TRUE)))
  if(length(hasopt)==0) return(x)
  hasopt <- grep("^\\s*(>>|@)", x[hasopt], perl = TRUE)
  if(length(hasopt)==0) return(x)
  x[-hasopt]
}

dump_opts(code)
#> [1] "root = \"@cppstem\"" "run = 1002"

dump_opts_old(code)
#> character(0)
```

<sup>Created on 2023-01-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/utils.R
kylebaron: The motivation was to avoid making two passes through the string since I wanted the extension as well. `file_ext()` will get the position of the `.` on the extension and if we could get access to that, it gives clean way to split the string. So I wrote that myself and then used the regmatch to split. I can change it to boxed code.

kylebaron: This is kind of weird, though:

```
> tools:::file_path_sans_ext(".foo")
[1] ".foo"
> tools:::file_ext(".foo")
[1] "foo"
```

kyleam: > The motivation was to avoid making two passes through the string since I wanted the extension as well.

Right, my confusion is that I'm not seeing where the extension is ever used (perhaps I'm missing that, or perhaps you want it because you expect to use it for some future feature?)

> I can change it to boxed code.

If you prefer the custom function, no objections from me.  I just wasn't sure of the motivation.

> This is kind of weird, though:

Hmm, yeah, that inconsistency is weird.  

I guess it doesn't really matter in this context, given a cppfile with a leading dot would be unlikely, but it seems like the `file_path_sans_ext` behavior is probably more appropriate, right?

R/modspec.R
kylebaron: Yes; I see it now. Will fix ... need to get some project work out right now. Thanks for picking it up.

R/utils.R
kylebaron: Ok; dropped the custom code and no longer tracking `ext`.  Dropped the test too. Will re-visit if / when I need to track `ext`. 

Thanks.

R/modspec.R
kylebaron: Removed the over-complexity in `dump_opts()` and added a test. Thanks.

kyleam (approved):
Thanks, looks good to me.

The PR drone build is failing due to a stories.yaml conflict with the PR base. Locally I merged `origin/develop` into this branch and confirmed that `make test-all` still passes, so should be all good once the conflict is resolved. Let me know if you'd like me to push that conflict resolution (i.e. the merge of `origin/develop` into this PR branch) to this PR.

PR #1037, opened 2022-12-20
User can provide ETAs on the data set

kylebaron:
# Summary

- rather than simulating new ETAs, the user adds `ETA1`, `ETA2`, ... `ETAn` as columns on the input data set

## Details

- pass `etasrc = "omega"` to simulate from omega (current default, previous behavior)
- pass `etasrc = "data"` to scan data set columns for `ETAn`; whatever is found is used to form the `nid` x `neta` matrix of `ETAs`; if an `ETAn` _isn't_ found, we'll fill that with `0`
- pass `etasrc = "data.all"` to function just like `etasrc = "data"` but generate an error if we don't find all `ETAn`
- When the `ETA` number is greater than `9`, start looking for `ETn` first; if we don't find that, then look for `ETAn`
- It is an error to have both `ETn` and `ETAn` in the data set for `n` > 9
- It is an error to pass `"data"` or `"data.all"` _and_ a data set with no `ETAn` present

See the attached pdf for detail and behavior. 
[etasrc.pdf](https://github.com/metrumresearchgroup/mrgsolve/files/10761364/etasrc.pdf)


FelicienLL:
Hi Kyle, 
very nice feature, I can already imagine many situation this would be helpful. 
I have one question, do you think it could be possible to set the model with default values of ETA(1) ETA(2) that would not be equal to zero? I'm asking because in mapbayr, you know that we run the trick where each ETA must be duplicated (ETA(1) for stochastic simulations, and ETA1 in $PARAM for deterministic predictions). It is working well but I'd love that we don't have to do this trick and if we could use a "regular" mrgsolve model.
With your proposal implementation (ETA_1 read from the dataset), it would now be possible to do some Bayesian estimation of ETA from a regular mrgsolve model (internally, I would just add ETA_1 ETA_2 variables in the dataset instead of updating ETA1 ETA2 in $PARAM as it is currently implemented). But once the estimation is done, the user can currently call `use_posterior()` in order to return an updated model with the estimation of ETA1, ETA2 in $PARAM (e.g. ETA1 = 0.123, ETA2 = -0.456), so that the user can simulate alternative dosing regimens for the patient, from a model object only, and no need to remember to pass individual values of ETA1, ETA2 in a data set. 
So, to sum it up, do you think it would be possible to attach a vector of ETA values to the model? I would be happy to discuss more details if you are interested in. 
Many thanks in advance
Félicien

kylebaron:
Hi @FelicienLL  - Yes ... I was definitely thinking of you and `mapbayr` when working up this feature. We have a different use case for needing option of either ETA (fixed or newly simulated) in the problem and want to come up with a way to handle all of those. 

Some questions:

1. I was thinking of allowing alternative workflow of letting the user attach a vector or matrix to draw ETA from
2. Should we require all ETAs to be present and just take them in order? or identify ETAs from the names?
3. Should ETA vector or matrix be "stored" in the model object? or have to be passed through `mrgsim()`?

Other questions coming; thanks for your input and discussion on this. 

Kyle

FelicienLL:
Something else currently implemented in mapbayr thanks to the `ETA1 + ETA(1)` trick is the possibility to let mrgsolve simulate from (a gaussian approximation of) the posterior distribution: MVN(**μ**,**Ω**) with **μ** the vector of estimates of ETA and **Ω** a variance-covariance matrix obtained from the hessian (like NONMEM provides in the .phi file). Indeed, what I do in `mapbayr` is that I update ETA1,ETA2 in `$PARAM` with **μ** and `$OMEGA` with **Ω** so that mrgsolve simulates `ETA1 + ETA(1)` with a distribution **μ** + MVN(0,**Ω**), which is, correct me if I'm wrong, equal to MVN(**μ**,**Ω**).

So, instead of implementing a vector of default ETA in the model as I initially suggested, maybe it would be more relevant to give the possibility to change the mean of the multivariate normal distribution ETA are drawn from? This single vector would be defaulted to `c(MU_1 = 0, MU_2 = 0)` but could be updated with other values, either at the estimation steps or for "posterior" stochastic simulations.

In addition with your proposal of passing ETA_1,ETA_2 from the data, we would deal with any possibilities: 
- Fixed individual parameters at the model level: `mapbayr::use_posterior()` would return a model with `c(MU_1 = 0.12, MU_2 = -0.34)`. This way, no need for the user to manually pass `ETA_1 = 0.12` in the data or to mrgsim. And if the user wants to run stochastic simulations from the var-cov derived from the hessian, we would just have to update `$OMEGA`
- Simulate multiple IDs from a pre-existing matrix of ETA (newly simulated or extracted from NONMEM's .phi): pass them as ETA_1,ETA_2 column in data_set/idata_set.

kylebaron:
Hi @FelicienLL - 

Just to follow up on this: really cool how you're tweaking ETA and OMEGA to do these simulations. I think there could be a way to carry this information around in the model object, but likely require some specialization of the model. Let's see how this current proposal works out; I'm hopeful it will still work for you to get the `ETA1 + ETA(1)` requirement out from `mapbayr`.  I don't know what the performance will be like having to update the data set every iteration; but willing to see about alternatives (like `idata` or an object in `$ENV` that is used to pass the information through). 

Kyle

FelicienLL:
Hi, this looks very good. I noted that you enabled the possibility to pass variables called ET11, ET12 instead of ETA11, ETA12, which would be helpful for ETAs coming from NONMEM's $TABLE. 
``` r
library(mrgsolve)
#> 
#> Attachement du package : 'mrgsolve'
#> L'objet suivant est masqué depuis 'package:stats':
#> 
#>     filter
code <- "
$OMEGA 1 1 1 1 1 1 1 1 1 1 1
$CAPTURE ETA(11); 
"
mod <- mcode("mod", code, end = -1)
#> Building mod ...
#> done.

data1 <- as_data_set(ev(amt = 1, cmt = 0, ETA11 = 99))
mrgsim_df(mod, data1, etasrc = "data")
#>   ID time ETA_11
#> 1  1    0     99

data2 <- as_data_set(ev(amt = 1, cmt = 0, ET11 = 9999))
mrgsim_df(mod, data2, etasrc = "data")
#>   ID time ETA_11
#> 1  1    0   9999

data2bis <- as_data_set(ev(amt = 1, cmt = 0, ET3 = 9999))
try(mrgsim_df(mod, data2bis, etasrc = "data"))
#> Error : at least one ETA must be provided when etasrc is data.
```

There are however no warnings if both syntax are present (ET11 has the priority over ETA11), is this something we should be concerned about? I acknowledge the case would probably be rare.

```r
data21 <- as_data_set(ev(amt = 1, cmt = 0, ET11 = 9999, ETA11 = 99))
mrgsim_df(mod, data21, etasrc = "data")
#>   ID time ETA_11
#> 1  1    0   9999

data12 <- as_data_set(ev(amt = 1, cmt = 0, ETA11 = 99, ET11 = 9999))
mrgsim_df(mod, data12, etasrc = "data")
#>   ID time ETA_11
#> 1  1    0   9999
```

And I was a bit confused that the syntax `ETA_11` was not allowed since it is the name used internally, and returned by `mrgsim()` itself. I could imagine the case of a user who want to make reproducible simulations by running mrgsim() first, capturing the ETA_n, and re-use them into a new dataset in order to be reproducible. Would you say it is something to be considered? 
```
data3 <- as_data_set(ev(amt = 1, cmt = 0, ETA_11 = 99))
try(mrgsim_df(mod, data3, etasrc = "data"))
#> Error : at least one ETA must be provided when etasrc is data.
```
Thanks 
Félicien

kylebaron:
Hi @FelicienLL - 

Thanks for reviewing this; your feedback is really important to me. 

1. I added a check for ambiguous names (like passing both `ETA11` and `ET11`); it is an error do do this. What do you think?
2. I want the default handling to accept what NONMEM tables out and agree it's a little inconsistent to have this not matching the default output from mrgsolve. I'm going to leave it as is for now; the names can get converted pretty easily and we could eventually provide a function ton convert them or even an `ETAS(1:last)` type notation that would use `ETAn` type naming. Maybe there's a block option `@eta-sep` that would let the user pick if or what separator would go there. 

``` r
library(mrgsolve)
#> 
#> 
#>     filter

code <- '
$OMEGA
1 1 1 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1 1 1

$CAPTURE ETA(1), ETA(9), ETA(10), ETA(16), ETA(20), ETA(21)
'

mod <- mcode("code", code)
#> Building code ...
#> done.

data <- expand.ev(amt = 100, ETA1 = 2, ETA5 = 5, ETA9 = 9, ET10 = 10, ET16 = 16, 
                  ETA20 = 20, ET21 = 21, ET20 = 202, cmt = 0)

mrgsim(mod, data, etasrc = "data")
#> Error:
#> ! Ambiguous ETA names when `etasrc` is "data":
#> ✖ Found both ETA20 and ET20 in the data.

#> Backtrace:
#>     ▆
#>  1. ├─mrgsolve::mrgsim(mod, data, etasrc = "data")
#>  2. │ └─mrgsolve::mrgsim_d(x, data = data, ...)
#>  3. │   ├─base::do.call(...)
#>  4. │   └─mrgsolve (local) `<fn>`(...)
#>  5. └─mrgsolve (local) `<fn>`(`<chr>`, 20L, "data")
#>  6.   └─rlang::abort(...)
```

<sup>Created on 2023-02-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

FelicienLL:
Hi Kyle, thanks for your feedback too.
1. Looks good to me. Errors well if multiple ambiguities.
2. Also agree that compatibility with NONMEM should be a priority. My suggestion for implementing `ETA_i` compatibility would be, when `etasrc = "data"`, to first look for values named either `ETA_i`, `ETAi` , or `ETi` , then check ambiguous names, and rename these into `ETA_i` internally. This way, mrgsolve still uses `ETA_i` internally, still outputs `ETA_i`, accepts data with `ETA_i`, and *also* accepts data with `ETAi` and `ETi` as aliases for `ETA_i`. And to be extremist, we could also imagine compatibility with `@labels` if they were defined in `$OMEGA` (looking for `ECL` in the data etc...).

Some more things came to my mind but I don't know if they are relevant at all, so feel free to ignore: 

3. For each ID, only the first row is looked at. Should mrgsolve check unique values in `ETAi` and error if several different values are found for a single individual?
4. From what I experienced, `ETAi` will not be looked for if it comes from `ev()` or `idata_set()`. Should it be stated in the documentation? Should this be implemented as a feature? Or nothing (would probably be fine enough, I don't know).

R/mrgsolve.R
```diff
@@ -18,6 +18,34 @@
...
+#' Check ETA names in the data set
+#' 
+#' This function checks to see if there are both `ETAn` and `ETn` in the data
+#' set when `n` is greater than 9. For example, when both `ETA12` and `ET12` 
+#' are found, an error is generated. This funciton is only use for its 
+#' side-effects. 
+#' @noRd
+check_etasrc_names <- function(data_names, neta, etasrc) {
+  eta_check <- paste0("ETA", seq(10, neta))
+  et_check <- paste0("ET", seq(10, neta))
```
kyleam, line 30: Mostly thinking aloud as reading:`dataobject::get_etas` only calls this when it's processing `i==10`, so the current code is never going to hit the `neta` less than 10 descending case (e.g., `seq(10,9) -> 10 9`). And even if somehow the calling code passed through `neta < 10`, it would just lead to wasted computation, not an error.

So, the current level of guarding seems sufficient for an internal function.

R/mrgsolve.R
```diff
@@ -18,6 +18,34 @@
 
 # @include mrgindata.R
 
+#' Check ETA names in the data set
+#' 
+#' This function checks to see if there are both `ETAn` and `ETn` in the data
+#' set when `n` is greater than 9. For example, when both `ETA12` and `ET12` 
+#' are found, an error is generated. This funciton is only use for its
```
kyleam, line 25: `s/funciton/function/`

`s/use/used/`

R/mrgsolve.R
```diff
@@ -127,6 +155,33 @@ tgrid_id <- function(col,idata) {
...
+#' mrgsolve finds both the `ETA` and `ET` name variant for the tenth and higher 
+#' `ETA` (e.g. it is an error to have both `ETA10` and `ET10` in the data set). 
+#' When mrgsolve is searching for `ETA` columns in the data set, it will 
+#' _only_ look for `ETAn` up to the number of rows (or columns) in all the 
+#' model `$OMEGA` blocks. For example, if `$OMEGA` is 5x5, only `ETA1` through 
+#' `ETA5` will be searched. An error will be generated in case mrgsolve finds 
+#' _no_ columns with `ETAn` names and something other than `etasrc = "omega"` 
+#' was passed. When `etasrc = "data"` and an `ETAn` column is missing from the 
+#' data set, the missing `ETA()` will be set to `0`.  Alternatively, the user 
+#' can pass `etasrc = "data.all"` which cause an error to be generated if any
```
kyleam, line 180: `s/which cause/which causes/` ?

R/mrgsolve.R
```diff
@@ -489,6 +547,10 @@ do_mrgsim <- function(x,
     x <- update_outputs(x, outputs = Request)  
   }
   
+  if(!is.character(etasrc) && length(etasrc)==1) {
```
kyleam, line 550: If I understand correctly, this is meant to catch anything that's not a character scalar.  So I guess the condition should be either `!(is.character(etasrc) && length(etasrc) == 1)` or `!is.character(etasrc) || length(etasrc) != 1`?

Examples of cases that I think should be caught but aren't, leading to an error message that's a little less helpful:

```
> mrgsim(mod, data, etasrc = c("data","foo"))
Error: Expecting a single string value: [type=character; extent=2].
> mrgsim(mod, data, etasrc = 1:2)
Error: Expecting a single string value: [type=integer; extent=2].
```

kyleam (changes requested):
I'm not very familiar with all of the different pieces here, but this looks good to me overall.  As usual, the context you've provided is very good at explaining the motivation/use case.

My main comment is that the "etasrc is string" condition looks off (though the fallout is just a more obscure error.)

R/mrgsolve.R
kylebaron: Thanks, @kyleam. Yes ... the logic I had wasn't right. Will update to your first proposal.

kylebaron: Both of these are passing now

``` r
library(mrgsolve)
#> 
#> 
#>     filter
code <- '
$OMEGA
1 1 1 1 1 1 1 1 1 1
1 1 1 1 1 1 1 1 1 1

$CAPTURE ETA(1), ETA(9), ETA(10), ETA(16), ETA(20), ETA(21)
'

mod <- mcode("code", code)
#> Building code ...
#> done.


data <- expand.ev(
  amt = 100, ETA1 = 2, ETA5 = 5, ETA9 = 9, ET10 = 10, ET16 = 16, 
  ETA20 = 20, ET21 = 21, cmt = 0
)

mrgsim(mod, data, etasrc = c("data", "foo"))
#> Error:
#> ! `etasrc` must be a string.

#> Backtrace:
#>     ▆
#>  1. └─mrgsolve::mrgsim(mod, data, etasrc = c("data", "foo"))
#>  2.   └─mrgsolve::mrgsim_d(x, data = data, ...) at m4solve/R/mrgsolve.R:272:6
#>  3.     ├─base::do.call(...) at m4solve/R/mrgsolve.R:349:2
#>  4.     └─mrgsolve (local) `<fn>`(...)
#>  5.       └─rlang::abort("`etasrc` must be a string.") at m4solve/R/mrgsolve.R:551:4

mrgsim(mod, data, etasrc = c(1,2))
#> Error:
#> ! `etasrc` must be a string.

#> Backtrace:
#>     ▆
#>  1. └─mrgsolve::mrgsim(mod, data, etasrc = c(1, 2))
#>  2.   └─mrgsolve::mrgsim_d(x, data = data, ...) at m4solve/R/mrgsolve.R:272:6
#>  3.     ├─base::do.call(...) at m4solve/R/mrgsolve.R:349:2
#>  4.     └─mrgsolve (local) `<fn>`(...)
#>  5.       └─rlang::abort("`etasrc` must be a string.") at m4solve/R/mrgsolve.R:551:4
```

<sup>Created on 2023-02-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/mrgsolve.R
kylebaron: Yes .... so the intention is that this _should_ only get called if there are more than 9 ETAs. So I think it should hold up as long as the behavior stays the same. 

What do you think about 

- generating an error in case `neta` <= 9 just for my own benefit in case the behavior changes?
- returning early in case `neta` is <= 9? just to reiterate that we don't need  this check in this case

Sort of opposites ... but wouldn't be against doing something like that just to protect me from my (future) self.

kyleam: I think adding a guard that errors for the `neta <= 9` case is a good idea: prevents future callers from making mistakes, as you mention, and also makes the intention clearer to the reader.

PR #1039, opened 2023-01-11
fixes gen rdefs issue for nm-vars with no frda

kylebaron:
# Summary

When processing a model with `nm-vars`, if there wasn't an `FRDA` item, we'd write out a bad pre-processor definition. This PR  implements check of `found_frda` rather than `found_any` when deciding whether or not to try to write a definition out. Also noticed that we were carrying `found_frda` in the object, but it was never set.

PR #1042, opened 2023-01-23
Release/1.0.7

kylebaron:
# mrgsolve 1.0.7

- `$NMXML` and `$NMEXT` now accept the `run` argument set to `"@cppstem"` (i.e  
  `run = "@cppstem"`); in this case, the stem of the NONMEM run will be assumed
  to be the same as the stem of the mrgsolve model file (#1025).

- Missing values (`NA`) in input data sets will be replaced with 0 for the
  following columns: `AMT`, `CMT`, `EVID`, `II`, `ADDL`, `RATE`, `SS` as well as
  their lower case counterparts (#1030).

- Refactored include order when building a model; this is an internal update and
  not expected to be visible to the user (#1038).

## Bugs Fixed

- Fix bug in generating certain model definitions when using the `nm-vars`
  plugin; the bug would have resulted in a warning from the pre-processor and
  did not affect function of the model (#1039).


kyleam (approved):
[release changes](https://github.com/metrumresearchgroup/mrgsolve/compare/2db48da5ae3371f5de4c39fb3a62f439c93262e0...ec6651e5beabc92c20f5414d7ab47681e08dc52c) look good to me

PR #1048, opened 2023-02-02
use identical to check value of run in nmext and nmxml

kylebaron:
# Summary

This PR fixes a bug in `NMEXT` and `NMXML` where the `run` argument may take it's default value of `NA` (nmext) or `NULL` (nmxml) and then be tested for equality with `"@cppstem"`. This bug was introduced in version 1.0.7 with the `"@cppstem"` feature. 

The test of `run` was updated to use `identical()` rather than `==`.  The PR also adds two unit test models where we bypass the `run` + `project` setup and specify `path` instead. The `NMEXT` test will fail if using version 1.0.7 but pass with this PR. I didn't check this with `NMXML`, but updated the test logic to `identical()` anyway and the test should pass with this PR. 

Some tests were modified to avoid actually compiling the models as compilation is not needed for these tests.

### DIFF
--- a/R/nmxml.R
+++ b/R/nmxml.R
@@ -102,7 +102,7 @@ nmxml <- function(run = numeric(0), project = character(0),
     if(missing(run) | missing(project)) {
       wstop("both file and run or project are missing")
     }
-    if(run=="@cppstem") {
+    if(identical(run, "@cppstem")) {
       run <- env$root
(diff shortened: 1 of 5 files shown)


kyleam (approved):
fix looks good, sorry for not spotting this case in the initial review

PR #1051, opened 2023-02-07
Pass SIGMA into model functions

kylebaron:
# Summary

This PR implements the ability to access on diagonal elements of `SIGMA` in the model functions. 

The data are stored as a numeric vector in `self.SIGMA`. This is zero- based indexing. There is 1-based indexing through the `SIGMA()` macro (see unit test).


timwaterhouse (approved):
Wow, you're fast. This looks good, thanks.

PR #1055, opened 2023-02-20
Bulk capture of ETAs

kylebaron:
This PR implements a feature where the user can capture ETA values with a concise syntax and where the names in the simulated output look like the names coming out of the NONMEM $TABLE step. In order to do this, some existing internal functionality had to be refactored as noted below. 

## Background

Along with this bulk capture feature, I've re-factored how `CAPTURE` data are accumulated during the model parse and load sequence.  Previously, we could only `CAPTURE` items through the `$CAPTURE` block in the model. Then, we added functionality to dynamically capture items .... that is, we can name model data items when calling `mread()` and the model will get compiled with all those items in the capture list, even the items that weren't written into the model file. This was implemented in a very straightforward way, gently extending what we had done before. Now that we want to capture `ETAs`, I felt like we needed to refactor this code to make it more robust and easier to maintain. 

Part of refactoring the `CAPTURE` process involves refactoring a lightweight class + methods for renaming things. This has been around for a while and has worked well. When the user lists variables to `CAPTURE`, they can write

```r
$ERROR
double IPRED = A(2)/VC;

$CAPTURE CONC = IPRED
```

so that the output will contain a column called `CONC` and it will take the value of `A(2)/VC` or the same as `IPRED`. When the capture variables are read in, they are parsed into an object that tracks the old name (`IPRED` in this example) and the new name (`CONC` in this example) as well as a flag indicating whether or not the old names are identical to the new names. There are several places in the simulation workflow where we allow renaming of data columns and they all use this functionality. 

Part of renaming items from `CAPTURE` is the sanitizing of the names. By this we simply mean removing opening and closing parens and brackets. So far, this mainly let's us `CAPTURE` things like `ETA(1)` and it will show up in the simulated output as `ETA_1`.  I still want to retain this behavior, but the PR also let's us get ETAs into the output in another way (see below). 

You will know this select / rename class (doesn't have a class attribute) and methods with the `.ren.` prefix (e.g. `.ren.create()` to create a new object and `.ren.chr()` to spit out an ordinary character vector with the new names as data and the old names as names attribute on the character vector. 

## Details

1. Start using `rlang::abort()` ... several cases where the extra flexibility comes in handy and makes sense to dive in with this now.
2. The `CAPTURE` block now comes with an `@etas` option. The user passes in the integer values of the ETAs to capture. The text that gets entered will get parsed and evaluated in environment where `last` and `LAST` (the total number of ETAs in the model) are available. The intention here is to mimic the syntax in NONMEM. So for example, you can specify `@etas 1:last` to get all the ETAs into the captured output. 
3. There is now a `update_capture()` function that is internal; this takes in new (additional) items to capture and adds these items into the model object. This is the primary refactor that allows us to incrementally find capture data and add it to what is already known.
4. Previously, there was ability to sanitize names that was embedded in the `.ren.create()` method. This functionality has been added as a separate step (`.ren.sanitize()`). Sanitize means removing parens or brackets out of variable names so that the user can capture `ETA(1)` and they will have `ETA_1` in the output. 
5. Change the default value of `sanitize` argument to `.ren.create()` to `NULL`; we will retain the ability to sanitize inside of the `.ren.create()` function for now with a `TODO` of dropping this. It's internal ....
6. An internal function `sanitize_capture` was refactored to run ` gsub()` on the vector of names rather than looping through the items to sanitize (it was not needed).
7. Tests added for the sanitize behavior as well as the directive in `$CAPTURE` to capture ETAs in bulk as NONMEM does


R/modspec.R
```diff
@@ -39,6 +39,68 @@ write_capture <- function(x) {
...
+    if(inherits(etan, "try-error")) {
+      msg <- c(
+        glue("could not parse this expression for `etas`: {eta_txt}."), 
+        x = etan
+      )
+      abort(msg)
+    }
+    resolves_int <- 
+      is.integer(etan) || 
+      (is.numeric(etan) && all(abs(etan - round(etan)) < 1e-08))
```
kyleam, line 80: Fine as is, though perhaps it'd be better to match the same default tolerance of `sqrt(.Machine$double.eps)` (normally 1.490116e-08) that's used by `all.equal.numeric` and `dplyr::near` and in the example section of `help(integer)`.

R/modspec.R
```diff
@@ -39,6 +39,68 @@ write_capture <- function(x) {
...
+    if(length(etan)==0) {
+      abort("`etas` has length 0.")  
+    }
+    etan <- unique(as.integer(round(etan)))
+    if(any(etan > last)) {
+      abort(
+        message = c(
+          glue("`etas` must be integers between 1 and {last}."),
+          i = glue("minimum value in `etas`: {min(etan)}"),
+          i = glue("maximum value in `etas`: {max(etan)}")
```
kyleam, line 93: Should this check the lower bound too (i.e. that no values are below 1)?

```diff
diff --git a/inst/maintenance/unit/test-capture.R b/inst/maintenance/unit/test-capture.R
index 90647e5f..7d04ea49 100644
--- a/inst/maintenance/unit/test-capture.R
+++ b/inst/maintenance/unit/test-capture.R
@@ -139,4 +139,10 @@ test_that("capture with @etas directive", {
     mcode("capture-at-etas-7", code), 
     regexp = "has length 0"
   )
+
+  code <- paste0(base, "-1:5")
+  expect_error(
+    mcode("capture-at-etas-8", code),
+    regexp = "must be integers between 1 and 5"
+  )
 })
```

R/utils.R
```diff
@@ -482,11 +482,9 @@ na2zero <- function(x) {
...
 }
 
-sanitize_capture <- function(x) {
-  for(i in seq_along(x)) {
-    x[i] <- gsub("\\[|\\(", "_", x[i])
-    x[i] <- gsub("\\]|\\)", "",  x[i])
-  }
+sanitize_capture <- function(x, sep = "_") {
+  x <- gsub("\\[|\\(", sep, x, perl = TRUE)
+  x <- gsub("\\]|\\)", "",  x, perl = TRUE)
```
kyleam, line 487: The `perl = TRUE` switch wasn't needed, right?  (Just making sure I'm not missing something.)

R/modspec.R
```diff
@@ -39,6 +39,68 @@ write_capture <- function(x) {
...
+#' `1:last` which will be integer, but we want to support `c(1,2,5)` as well
+#' which will not be integer. 
+#'
+#' @return The model object, possibly updated.   
+#' 
+#' @noRd
+capture_etas <- function(x, env) {
+  if(!is.character(env[["capture_etas"]])) return(x)
+  last <- dim_matlist(omat(x))
+  if(last==0) return(x)
```
kyleam, line 66: It looks like `dim_matlist` uses `sapply`, so does this `if` condition need to worry about handling a non-scalar `last`?

kyleam (approved):
Looks like a nice feature, and the background and details make sense to me.  Left a few minor comments/suggestions.

R/modspec.R
kylebaron: Thanks. I didn't think about `all.equal` or `near()`. Will switch to one of those. I was looking at `is_integerish()` which is what I really want but wasn't sure about the maturity of that one.

R/modspec.R
kylebaron: Thanks; added the unit test and updated the check in the code.

R/utils.R
kylebaron: Right; wasn't needed.

R/modspec.R
kylebaron: Thanks. This would have been a disaster. I thought I had a method to give the total number of rows for OMEGA but I guess not. For some reason created this `dim_matlist()` function when I already had an `nrow()` method that gives the same result. Thanks again for finding this!

kylebaron: Also updated the test to include multiple OMEGA blocks.

Issue #1050, opened 2023-02-06
mread cache gets the project folder wrong when the full path is passed


kylebaron:
When you pass the full path to a model file to `mread_cache()`, it crumps. 

```r
> mod <- mread_cache("../model/sim/207.mod", capture = "CLI")
Error: project directory '__/model/sim' must exist and be readable.
> mod <- mread_cache("207.mod", project = "../model/sim", capture = "CLI")
Building 207_mod ... done.
```

kylebaron:
Reproducible in package

```r
> x <- mread_cache("../m4solve/inst/models/1005.cpp")
Error: project directory '__/m4solve/inst/models' must exist and be readable.
```

kylebaron:
Fixed now

```r
> x <- mread_cache("../m4solve/inst/models/1005.cpp")
Loading required namespace: xml2
Building 1005_cpp ... done.
```

PR #1056, opened 2023-02-23
Fix mread_cache when the full path is passed

kylebaron:
See #1050 

This PR adds a bit of code that was already in `mread()` for dealing with this input.

A test was added for this scenario as well.

### DIFF
--- a/R/mread.R
+++ b/R/mread.R
@@ -159,7 +159,7 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
                   capture = NULL,
                   preclean = FALSE, recover = FALSE, ...) {
   
-  if(charthere(model, "/")) {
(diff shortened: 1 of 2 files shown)


R/mread.R
```diff
@@ -600,6 +600,11 @@ mread_cache <- function(model = NULL,
                         preclean = FALSE, 
                         capture = NULL, ...) {
   
+  if (charthere(model, "/")) {
```
kyleam, line 603: I'm not sure where mrgsolve stands with Windows compatibility, but this would fail if someone on Windows passed "\\", right?  

As an alternative check/condition, the first thing that comes to mind is

```
!identical(basename(model), as.character(model))
```

(with `as.character` there in case someone passes in an `fs_path` object)

kyleam (approved):

PR #1060, opened 2023-02-28
Drop cxx11

kylebaron:
This PR drops any statement of a C++ standard. This seems to be required 
See: https://cran.r-project.org/web/checks/check_results_mrgsolve.html

A lot of people are saying to just drop it. I'd personally rather just leave it in but it looks like CRAN is going to complain if it finds `CXX11`. 

One question I have: should we start depending on a later version of R? Maybe it's time to bump that up. I don't think there is much to worry about wrt CXX version if R >= 4.0.

https://developer.r-project.org/blosxom.cgi/R-devel/2023/02/21#n2023-02-21
https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#C-standards

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -66,7 +66,6 @@ Language: en-US
 LazyLoad: yes
 NeedsCompilation: yes
 RoxygenNote: 7.2.3
-SystemRequirements: C++14
 Collate: 
     'RcppExports.R'
     'utils.R'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ URL: https://github.com/metrumresearchgroup/mrgsolve
 BugReports: 
     https://github.com/metrumresearchgroup/mrgsolve/issues
 Depends:
-    R (>= 3.5),
+    R (>= 3.6.2),
     methods
 Imports:
     Rcpp (>= 1.0.7),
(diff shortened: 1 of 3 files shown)


kyleam (approved):
Based on the links you gave and some additional poking around (e.g., [this r-package-devel thread](https://stat.ethz.ch/pipermail/r-package-devel/2023q1/008886.html)), this change makes sense to me.

IIUC this means that mrgsolve will fail to build for R versions before 3.6.2 (or actually before 3.6.0, unless on Windows) unless the user overrides the default standard.

From R 3.6.2 NEWS:

```
 Windows:

   • The default standard for C++ in package installation is C++11 (as
     it has been on other platforms where available since R 3.6.0: the
     default toolchain on Windows was defaulting to C++98).
```

Given that...

> One question I have: should we start depending on a later version of R? Maybe it's time to bump that up. I don't think there is much to worry about wrt CXX version if R >= 4.0.

... bumping the minimum to at least 3.6.2 seems reasonable to me, though I'm not sure what your policy/approach has been for supporting older R versions.

Drone's oldest CI build is with R 3.6.3, which gives at least some confidence that 3.6.x is still building.

kylebaron:
Thanks @kyleam; I did end up. bumping to >= 3.6.2. Good suggestion.

Issue #1052, opened 2023-02-07
Access SIGMA(n) in model functions


kylebaron:
# Summary
As a user, I want to access `SIGMA` elements inside model functions. 

For example: 

```c

$OMEGA
2.5
0.1

$ERROR
double STD=sqrt(SIGMA(1)+pow(F,2)*SIGMA(2));
```

Issue #1046, opened 2023-02-01
$NMEXT (and $NMXML) fail when run argument isn't provided


kylebaron:
Need some protective logic around `run`

```r
> mod <- mread("tests/testthat/nm/1005-path.mod")
 Error in if (run == "@cppstem") { : missing value where TRUE/FALSE needed 
> packageVersion("mrgsolve")
[1] ‘1.0.7.9000’
```

PR #1061, opened 2023-03-01
Release 1.0.8

kylebaron:
# mrgsolve 1.0.8

- `SIGMA()` is a new model macro which allows users to access on-diagonal
  elements of `SIGMA` in the model (e.g. `SIGMA(2)` in `$ERROR`) (#1051, #1052).
  
- `mrgsim()` and `mrgsim_q()` gain an `etasrc` argument, allowing `ETAs` to 
  be either simulated from `OMEGA` (new default and previously the only 
  behavior) or taken from the input data set (new option), similar to the way 
  parameters can be scraped from the data set (#1037).

- `@etas` is a new option for use with the `$CAPTURE` block to let users name
  `ETAs` to be captured into the simulated output; for example, use 
  `@etas 1:last` to capture all model `ETAs` in the simulated output (#1055).

- Drop `CXX_STD` statement from Makevars file and DESCRIPTION to be consistent
  with current changes in R-devel; mrgsolve continues to require compiler
  capable of implementing C++11 standard, but this should be selected
  automatically by R (#1060).

- mrgsolve now depends on `R >= 3.6.2` (#1060).

## Bugs Fixed

- Fix bug when the `path` argument is used in the `$NMXML` or `$NMEXT` blocks;
  this bug was introduced through the `@cppstem` feature in version `1.0.7` 
  (#1046, #1048).

- Fix bug in `mread_cache()` where the `project` directory wasn't getting 
  rendered properly when passing the complete path to the model specification
  file (#1056).


kyleam:
image

Shouldn't this be `release/1.0.8`?

`develop...release/1.0.8`: https://github.com/metrumresearchgroup/mrgsolve/compare/2ea99b856598829e9ecd3ae40b2320d3e2b5f440...d8f4e0d1e9cbbfbc9f6f6fb66d1f1891a95405ad

PR #1065, opened 2023-03-08
Fix generic signature problems identified with new R-devel

kylebaron:
# Summary

CRAN has come up with some methods where the signature does not match the generic. This has been there for a long time and I guess R-devel changed a bit to detect this; it isn't there in R-release or R-devel up to a couple of days ago. 

@kyleam - I'm not worried about the `compile()` method. But I really want to be able to have an `as_tibble()` method for the simulation output class. I had the signature right at some point but the generic changed and that caused the problems. 

I'm heading toward just getting the methods matched up with the generics and sending that back. That seemed to fix the issue when testing R-devel from yesterday. Wondering about your thoughts on safe / sustainable approach going forward?

# Results 

```
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_mrgsolve.html>.

Please correct before 2023-03-22 to safely retain your package on CRAN.

Best,
-k
```

https://cran.r-project.org/web/checks/check_results_mrgsolve.html

# Tests

## Reproduce check issue on `1.0.8`

I've only been able to reproduce on (very) recent R-devel.

```
R CMD check mrgsolve_1.0.8.tar.gz
* using log directory ‘/Users/kyleb/git/m4solve/mrgsolve.Rcheck’
* using R Under development (unstable) (2023-03-07 r83950)
* using platform: aarch64-apple-darwin20 (64-bit)
* R was compiled by
    Apple clang version 13.0.0 (clang-1300.0.29.3)
    GNU Fortran (GCC) 12.0.1 20220312 (experimental)
* running under: macOS Ventura 13.2
...
* checking dependencies in R code ... OK
* checking S3 generic/method consistency ... WARNING
compiled:
  function(x, ...)
compiled.mrgmod:
  function(x, status)

as_tibble:
  function(x, ..., .rows, .name_repair, rownames)
as_tibble.mrgsims:
  function(.data_, ...)

See section ‘Generic functions and methods’ in the ‘Writing R
Extensions’ manual.
* checking replacement functions ... OK
* checking foreign function calls ... OK

```

The changes in this PR seem to fix the issue 

```
 R CMD check mrgsolve_1.0.8.9001.tar.gz
* using log directory ‘/Users/kyleb/git/m4solve/mrgsolve.Rcheck’
* using R Under development (unstable) (2023-03-07 r83950)
* using platform: aarch64-apple-darwin20 (64-bit)
* R was compiled by
    Apple clang version 13.0.0 (clang-1300.0.29.3)
    GNU Fortran (GCC) 12.0.1 20220312 (experimental)
* running under: macOS Ventura 13.2
...
* checking dependencies in R code ... OK
* checking S3 generic/method consistency ... OK
*
```

### DIFF
--- a/R/mrgsims.R
+++ b/R/mrgsims.R
@@ -235,29 +235,18 @@ slice.mrgsims <- function(.data,...) {
 }
 
 ##' @rdname mrgsims_dplyr
-##' @param .data_ mrgsims object
 ##' @export
-as_data_frame.mrgsims <- function(.data_,...) {
-  as_tibble(as.data.frame(.data_),...)
+as_data_frame.mrgsims <- function(x,...) {
+  as_tibble(as.data.frame(x),...)
 }
 
-#' @param .rows passed to [tibble::as_tibble()]
-#' @param .name_repair passed to [tibble::as_tibble()]
-#' @param rownames passed to [tibble::as_tibble()]
+
+#' @param x mrgsims object.
+#' 
 #' @rdname mrgsims_dplyr
 #' @export
-as_tibble.mrgsims <- function(x, 
-                              .rows = NULL, 
-                              .name_repair = c("check_unique", "unique", 
-                                               "universal", "minimal"), 
-                              rownames = NULL, ...) {
-  as_tibble(
-    as.data.frame(x), 
-    .rows = .rows, 
-    .name_repair = .name_repair, 
-    rownames = rownames, 
-    ...
-  )  
+as_tibble.mrgsims <- function(x, ...) {
+  as_tibble(as.data.frame(x), ...)  
 }
 
 #' @rdname mrgsims_dplyr
--- a/R/mrgsims.R
+++ b/R/mrgsims.R
@@ -241,10 +241,23 @@ as_data_frame.mrgsims <- function(.data_,...) {
   as_tibble(as.data.frame(.data_),...)
 }
 
-##' @rdname mrgsims_dplyr
-##' @export
-as_tibble.mrgsims <- function(.data_,...) {
-  as_tibble(as.data.frame(.data_),...)  
+#' @param .rows passed to [tibble::as_tibble()]
+#' @param .name_repair passed to [tibble::as_tibble()]
+#' @param rownames passed to [tibble::as_tibble()]
+#' @rdname mrgsims_dplyr
+#' @export
+as_tibble.mrgsims <- function(x, 
+                              .rows = NULL, 
+                              .name_repair = c("check_unique", "unique", 
+                                               "universal", "minimal"), 
+                              rownames = NULL, ...) {
+  as_tibble(
+    as.data.frame(x), 
+    .rows = .rows, 
+    .name_repair = .name_repair, 
+    rownames = rownames, 
+    ...
+  )  
 }
 
 #' @rdname mrgsims_dplyr
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: mrgsolve
 Title: Simulate from ODE-Based Models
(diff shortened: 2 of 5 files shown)


R/mrgsims.R
```diff
@@ -241,10 +241,23 @@ as_data_frame.mrgsims <- function(.data_,...) {
...
+                                               "universal", "minimal"), 
+                              rownames = NULL, ...) {
+  as_tibble(
+    as.data.frame(x), 
+    .rows = .rows, 
+    .name_repair = .name_repair, 
+    rownames = rownames, 
+    ...
+  )  
 }
```
kyleam, line 261: Was it just unhappy about the name of the first parameter?  I can get the warning to go away on my end witth just `(x, ...)` as the signature.

kyleam (commented):
Hmm, I wonder if this is the result of [R's revision 83941](https://github.com/wch/r-source/commit/306cbd5d131ed8206f3a430ec1a27cd9e7fd392e).  Anyway I can trigger the warnings too with R-devel 2023-03-07 r83950 and mrgsolve 1.0.8.  I think the `as_tibble` does more than it needs to.

R/mrgsims.R
kylebaron: Thanks for testing that out. I thought methods had to have all the arguments of the generic. 

https://rstudio.github.io/r-manuals/r-exts/Generic-functions-and-methods.html

I could have sworn I ran into this before ...

kylebaron: @kyleam I went with the simpler fix as you suggested. I don't know what the status is of the `as_data_frame()` generic ... so I changed that argument too. I don't care if that stays or goes at this point so I just changed it to try to stay out of trouble. 


```r
> dplyr::as_data_frame
function(x, ...) {
  deprecate_warn("2.0.0", "as_data_frame()", "as_tibble()",
    details = "The signature and semantics have changed, see `?as_tibble`."
  )

  as_tibble(x, ...)
}
```

I sent this to R-devel to make sure it passed there. Will update when that comes back.

kyleam: > I thought methods had to have all the arguments of the generic.
>
> https://rstudio.github.io/r-manuals/r-exts/Generic-functions-and-methods.html

My understanding is that, although a method needs to accept all the arguments of the generic, it's okay for a method to fold arguments into `...`.  In the link above, using `...` in this scenario is referenced a bit in the `predict.lm` problem that's mentioned: "as `predict.lm` had neither a `dispersion` nor a `...` argument, `NextMethod` could no longer be used".

Note that some of tibble's own methods use `(x, ...)`:

```
tibble $ git describe HEAD
v3.2.0.9000

tibble $ git grep '^as_tibble\.' 'R/*.R'
R/as_tibble.R:as_tibble.data.frame <- function(x, validate = NULL, ...,
R/as_tibble.R:as_tibble.list <- function(x, validate = NULL, ..., .rows = NULL,
R/as_tibble.R:as_tibble.matrix <- function(x, ..., validate = NULL, .name_repair = NULL) {
R/as_tibble.R:as_tibble.poly <- function(x, ...) {
R/as_tibble.R:as_tibble.ts <- function(x, ..., .name_repair = "minimal") {
R/as_tibble.R:as_tibble.table <- function(x, `_n` = "n", ..., n = `_n`, .name_repair = "check_unique") {
R/as_tibble.R:as_tibble.NULL <- function(x, ...) {
R/as_tibble.R:as_tibble.default <- function(x, ...) {
```

kyleam (approved):

R/mrgsims.R
kylebaron: Great. This really simplifies things. I'll merge this and then send it to CRAN; the winbuilder (failed on 1.0.8) came back fine now.

PR #1068, opened 2023-03-21
Refactor/update dots

kylebaron:
# Summary

This pull request refactors bad behavior that was in place when passing arguments to `mrgsolve::update()`.  There was some bad reasoning behind the change introduced 4 years ago or so. It needs to be changed. 

By default, there should be at least a warning issued when an argument is passed to `mrgsolve::update()` that the function doesn't recognize. Eventually, I think this _should_ be an error, but historically it has been a warning so we'll keep it that way for now. 

There was a (bad) option controlling this; the option was bad and the implementation was bad. So this PR deprecates the option. This is handled in the documentation where we describe package-wide options and there is a warning issued if the user passes a bad argument to `mrgsolve::update()` with the option set. I highly doubt anyone is using this option so I don't anticipate there will be downstream effects. But we'll warn if the option is used _when the situation is right_. 

There is also an argument to `mrgsolve::update()` that controls if the warning should be issued when a bad argument is passed. This is really old, from when everything was organized quite different and we didn't have a very well defined model specification file. I think we might eventually drop that as well ... there is no need to make this behavior optional. So I wouldn't be against also deprecating the argument ... probably would be fine to just rip the band-aid off. But haven't done that (yet) in this PR.


kyleam (approved):
> By default, there should be at least a warning issued when an argument is passed to `mrgsolve::update()` that the function doesn't recognize. Eventually, I think this _should_ be an error, but historically it has been a warning so we'll keep it that way for now.

Makes sense, and the code changes look good to me.

One comment about something that's not new to this PR but related: shouldn't `update`s description for `strict` say "warning" rather than "error"?

https://github.com/metrumresearchgroup/mrgsolve/blob/35ac94e11ca94d62d5d82bce19edf067c27a2bd3/R/update.R#L43-L44

It looks like it did error when that description was introduced in 33eab8f2 (2019-09-29) and then the description went stale in the next commit when 88c76c9c (warn when bad update is attempted from simulation function, 2019-09-29) demoted the error to a warning.

kylebaron:
Thanks @kyleam ; i'll update that and put in a TODO to finally clean this up.

PR #1069, opened 2023-03-27
Recode NONMEM-based model in internal model library

kylebaron:
# Summary

Recoding this model so that it uses the traditional model specification syntax.


timwaterhouse (commented):
Looks good. I'm just a bit confused by your comment about this model being coded in `nm-like`. Seems like `nm-like` is a completely different model?

kylebaron:
Thanks; you're right it is a different model. I had that tangled up in my head.

PR #1072, opened 2023-04-07
Add & syntax for ev_rx

kylebaron:
# Summary 

- Use `&` to indicate that multiple events should be done at the same time
- `perl = TRUE` not required; I just like perl

### DIFF
--- a/tests/testthat/test-ev_rx.R
+++ b/tests/testthat/test-ev_rx.R
@@ -43,7 +43,7 @@ test_that("parse dose only - infusion", {
   a <- ev_rx("100 over 2")
   b <- ev(amt = 100, rate = 100/2)
   expect_identical(a,b)
-
+  
   a <- ev_rx("100 ov 2")
   expect_identical(a,b)
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-ev_rx.R
```diff
@@ -107,4 +107,15 @@ test_that("after parameter can be decimal", {
...
 
-
+test_that("two events at the same time", {
+  a <- ev_rx("100 in 1 & 200 in 2 after 4")  
+  b <- c(ev(amt = 100, cmt = 1), ev(amt = 200, cmt = 2, time = 4))
+  expect_identical(a,b)
+  
+  a <- ev_rx("100 in 1 & 200 in 2 then 300 in 3 after 10")  
+  b <- c(ev(amt = 100, cmt = 1), ev(amt = 200, cmt = 2), 
+         ev(amt = 300, cmt = 3, time = 10))
```
kyleam, line 117: Minor suggestion: I think making `b` completely reflect the expected output would be a bit nicer.  If you added `addl = 0, ii = 0` to all of these `ev` calls, you could just use `expect_identical(a, b)`, right?

kyleam (approved):
Looks good to me.

tests/testthat/test-ev_rx.R
kylebaron: Good idea; done.

PR #1076, opened 2023-04-07
allow parameter_list objects to be passed to param

kylebaron:
See discussion in #1075 : a user tried to do this in code and it didn't work; no reason why it shouldn't work.

This PR enables the user to pass a `parameter_list` object  to update parameters in a model object via `param()`

```r
library(dplyr)
library(mrgsolve)

mod <- house()

mod %>% param(param(CL = 5)) %>% param()
```


kyleam (approved):
Makes sense to me.

Confirmed that the new test fails when the code change is reverted.

Issue #1080, opened 2023-04-13
Documentation for mrgsim() argument 'nid' possibly misleading.


bergsmat:
Hi Kyle.

For ```?mrgsim```, (mrgsolve_1.0.6) I see "nid: integer number of individuals to simulate; only used if idata and data are missing".  But in the function body, the first logic branch seems to be 
```
if (nid > 1) {
        return(mrgsim_nid(x, nid, events, ...))
    }
```
So perhaps the doc should say "only used if greater than 1 (idata and data will be ignored)" or some such.  Quite possible that I have misunderstood!

PR #1086, opened 2023-05-03
Refactor mrgsim_nid

kylebaron:
# Summary

`mrgsim_nid()` is a convenience variant to simulate a number of IDs. The `nid` simulation has been around since way back when this was sort of useful. Now just keeping it around for continuity.

A user called out inaccurate documentation around what happens when `nid` is set to 1. Rather than changing the documentation, I refactored how it works. The bottom line is no meaningful change in function.

See #1080 

1. Change default `nid`  from 1 to `NULL`
2. Always call `mrgsim_nid()` when `nid` is numeric

### DIFF
--- a/R/mrgsolve.R
+++ b/R/mrgsolve.R
@@ -259,7 +259,7 @@ mrgsim <-  function(x, data=NULL, idata=NULL, events=NULL, nid=NULL, ...) {
   x@args$data <- NULL
   x@args$events <- NULL
   
-  if(is.numeric(nid) && !have_idata) {
+  if(is.numeric(nid) && !have_idata && !have_data) {
     return(mrgsim_nid(x, nid, events, ...))
   } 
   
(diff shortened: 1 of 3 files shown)


R/mrgsolve.R
```diff
@@ -259,7 +259,7 @@ mrgsim <-  function(x, data=NULL, idata=NULL, events=NULL, nid=1, ...) {
   x@args$data <- NULL
   x@args$events <- NULL
   
-  if(nid > 1) {
+  if(is.numeric(nid)) {
```
kyleam, line 262: The doc description is "[...] only used if idata and data are missing", so should the condition be updated to something like `!(have_data || have_idata) && is.numeric(nid)`?

kylebaron: Thanks; I think we're consistent with the doc now.

kyleam (approved):

PR #1088, opened 2023-05-09
Refactor interrupt interval checking

kylebaron:
This PR refactors how we tell when to check for interrupt signal. Previously, it used modulus; the refactor implements a clever trick that has recently been adopted in a bunch of R code. 

For example here: 
https://github.com/wch/r-source/commit/b5e5f267a6a16c679a1d186391b7d67b5ae76968

and here:
https://github.com/wch/r-source/commit/1ca6c6c6246629c6a98a526a2906595e5cfcd45e

Thread on R-devel
https://stat.ethz.ch/pipermail/r-devel/2023-May/082607.html

### DIFF
--- a/src/devtran.cpp
+++ b/src/devtran.cpp
@@ -435,7 +435,7 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
     
     for(size_t j=0; j < a[i].size(); ++j) {
       
-      if(do_interrupt && !(--ic)) {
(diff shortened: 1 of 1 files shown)


kyleam (approved):
Nice, thanks for links to external code and discussion.

Code changes look good to me, and I verified that I could still interrupt `mrgsim()` after this change.

PR #1092, opened 2023-07-12
Add support for etas attached to idata

kylebaron:
Add on to #1037 ; `etasrc` can now be
- `idata`
- `idata.all`

Re-using all of the `get_etas()` code. 

Checklist

- [x] Update devtran to accept `etasrc = "idata"` or `etasrc = "idata.all"`
- [x] Make sure `get_etas()` works with `idata` object; one minor change to fill out `Startrow` for idata sets.
- [x] Add tests
- [x] Documentation

 Discovered a bug when implementing #1092 where ETA was missed if it was in the first column of the data set. Please see #1095 for the fix.

### DIFF
--- a/src/dataobject.cpp
+++ b/src/dataobject.cpp
@@ -634,7 +634,7 @@ arma::mat dataobject::get_etas(const int n_eta, const bool strict,
   
   arma::mat ans(Uid.size(), n_eta); 
   for(size_t i = 0; i < Uid.size(); ++i) {
-    int start_row = Startrow[i];
+    int start_row = this->start(i);
     for(int j = 0; j < n_eta; ++j) {
       if(eta_location[j] >= 0) {
         ans(i, j) = Data(start_row, eta_location[j]);
--- a/src/dataobject.cpp
+++ b/src/dataobject.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2013 - 2019  Metrum Research Group
+// Copyright (C) 2013 - 2023  Metrum Research Group
 //
 // This file is part of mrgsolve.
 //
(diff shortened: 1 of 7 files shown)


src/dataobject.cpp
```diff
@@ -633,14 +634,13 @@ arma::mat dataobject::get_etas(const int n_eta, const bool strict,
   
   arma::mat ans(Uid.size(), n_eta); 
   for(size_t i = 0; i < Uid.size(); ++i) {
-    int start_row = Startrow[i];
+    int start_row = this->start(i);
     for(int j = 0; j < n_eta; ++j) {
-      if(eta_location[j] > 0) {
+      if(eta_location[j] >= 0) {
```
kyleam, line 639: I was confused at why this would be necessary for the `idata` change.  I reverted it (on top of c64ac9ab) and tests still pass.

Looking more closely at the existing upstream code, `eta_location` is the result of `std::find` on the eta labels, so 0-based.  By not considering `j = 0`, it mishandles an eta column if it's the first column.  So this is relevant for `data` too and is an independent fix for bug from gh-1037.  Okay.

Fwiw here's a regression test that fails without the `>` to `>=` equal change:

```R
test_that("etasrc works with ETA in first column", {
  mod <- param(mod, mode = 0)

  data <- expand.ev(amt = 100, ID = seq(4), cmt = 1)
  data <- mutate(data, ETA1 = rev(ID) / 10)
  data <- expand_observations(data, times = seq(5))
  data <- mutate(data, cmt = 0)

  set.seed(9812)
  expect_identical(
    mrgsim(mod, data, etasrc = "data"),
    mrgsim(mod, select(data, "ETA1", everything()), etasrc = "data")
  )
})
```

kyleam (approved):
> Re-using all of the get_etas() code.

That worked out cleanly.  Looks good to me.

src/dataobject.cpp
kylebaron: Thanks; yes, I discovered this when creating the tests for this PR. I should have created another PR or at least written it up in the message. I'm going to revert this change and make it clean in a new PR in case this comes up in the future.

FelicienLL (approved):
Thanks Kyle for providing this so fast. It should make things much easier and flexible. The code looks good to me.

Issue #1094, opened 2023-07-16
etasrc fails to get eta that is in the first column of data set


kylebaron:
This was discovered when working on  #1092 

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

mod <- modlib("popex", capture = "ECL,EV,EKA", end = 1)
#> Building popex ...
#> done.
data <- data.frame(ETA1 = 1, amt = 100, evid = 1, time = 0, cmt = 1, ID = 1)
data
#>   ETA1 amt evid time cmt ID
#> 1    1 100    1    0   1  1
mrgsim(mod, data, etasrc = "data") %>% select(ID, ECL)
#> # A tibble: 4 × 2
#>      ID   ECL
#>   <dbl> <dbl>
#> 1     1     0
#> 2     1     0
#> 3     1     0
#> 4     1     0

data <- data.frame(amt = 100, evid = 1, time = 0, cmt = 1, ID = 1,ETA1 = 1)
data
#>   amt evid time cmt ID ETA1
#> 1 100    1    0   1  1    1
mrgsim(mod, data, etasrc = "data") %>% select(ID, ECL)
#> # A tibble: 4 × 2
#>      ID   ECL
#>   <dbl> <dbl>
#> 1     1     1
#> 2     1     1
#> 3     1     1
#> 4     1     1
```

<sup>Created on 2023-07-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kylebaron:
Regression test suggested by @kyleam 


```r
test_that("etasrc works with ETA in first column", {
  mod <- param(mod, mode = 0)

  data <- expand.ev(amt = 100, ID = seq(4), cmt = 1)
  data <- mutate(data, ETA1 = rev(ID) / 10)
  data <- expand_observations(data, times = seq(5))
  data <- mutate(data, cmt = 0)

  set.seed(9812)
  expect_identical(
    mrgsim(mod, data, etasrc = "data"),
    mrgsim(mod, select(data, "ETA1", everything()), etasrc = "data")
  )
})
```

kylebaron:
[...]

PR #1095, opened 2023-07-16
Fix bug when ETA in first column when using etasrc="data"

kylebaron:
This bug was discovered when implementing #1092 ; when there is an ETA in the first column, it is not detected by `get_etas()`.

See #1094 and #1092 for discussion

### DIFF
--- a/inst/maintenance/unit/test-simeta.R
+++ b/inst/maintenance/unit/test-simeta.R
@@ -209,6 +209,22 @@ test_that("warn when simeps(n) is called with off diagonals", {
   expect_silent(mcode("simeps-n-nowarn-2", code, compile = FALSE))
 })
 
+# Fixes issue discovered in #1092
+test_that("etasrc works with ETA in first column", {
+  mod <- param(mod, mode = 0)
+  
+  data <- expand.ev(amt = 100, ID = seq(4), cmt = 1)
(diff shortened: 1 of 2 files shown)


inst/maintenance/unit/test-simeta.R
```diff
@@ -209,6 +209,23 @@ test_that("warn when simeps(n) is called with off diagonals", {
...
+  
+  data <- expand.ev(amt = 100, ID = seq(4), cmt = 1)
+  data <- dplyr::mutate(data, ETA1 = rev(ID) / 10)
+  data <- expand_observations(data, times = seq(5))
+  data <- dplyr::mutate(data, cmt = 0)
+  
+  set.seed(9812)
+  expect_identical(
+    mrgsim(mod, data, etasrc = "data"),
+    mrgsim(mod, dplyr::select(data, "ETA1", tidyselect::everything()),
```
kyleam, line 224: Interesting that the R 3.6 (but not R 4.0 or R 4.1) drone build [failed](https://github-drone.metrumrg.com/metrumresearchgroup/mrgsolve/1494/1/5) without f0354c3a given the pre-existing `library(mrgsolve)` at the top of the file.

kyleam (approved):
Thanks, looks good to me.

Issue #1100, opened 2023-08-05
mread() can't find project directory when working on a mapped network drive


marianklose:
Hi, thanks a lot for the great work! 👍 

I am having trouble with reading in model files when working on a mapped network drive. When working remotely we usually connect to the university network (e.g. \\\campus.fu-berlin.de\daten\...) via mapping it to a network drive (G://). It's hard to provide a reproducible example but here is the behaviour in a nutshell. 

To make sure that we are already in the model directory I run:

```
setwd("G:/Projekte/xxx/models")

getwd()
[1] "G:/Projekte/xxx/models"
```

Now there is a file in our working directory called model.cpp which is accessible via e.g., readLines()

```
readLines("goti2018.cpp")
  [1] "$PROB "                                                                                                      
  [2] " - Goti 2018 Model"                                                                                          
  [3] " - DOI: 10.1097/FTD.0000000000000490"                                                                        
  [4] " - Structure: two compartment"   
```     

However, mread() throws an error:

```
mread("goti2018")
Error: project directory 'G:/Projekte/xxx/models' must exist and be readable.
```

This is the traceback:

```
> traceback()
3: stop("project directory '", project, "' must exist and be readable.", 
       call. = FALSE)
2: new_build(file = file, model = model, project = project, soloc = soloc, 
       code = code, preclean = preclean, udll = udll, recover = recover)
1: mread("goti2018")
```

Inside new_build() which gets called by mread we have

```
if (!file_readable(project)) {
  stop("project directory '", project, "' must exist and be readable.", 
    call. = FALSE)
}
```

which apparently causes the error. The project argument should be the default argument from mread which is `project = getOption("mrgsolve.project", getwd())` and which is in our case "G:/Projekte/xxx/models". The definition of file_readable() from utils.R is

```
file_readable <- function(x) {
 file.access(x,4)==0 
}
```

Since file_readable isn't being exported I directly use file.access to check the behaviour:

```
> file.access("G:/Projekte/xxx/models", 4) == 0
G:/Projekte/xxx/models 
 FALSE 
```

The project is existent as `file.access("G:/Projekte/xxx/models", 0)` is TRUE. If I copy the full folder over to my Desktop both mread() and file access work without problems.

```
> file.access("C:/Users/Admin/Desktop/models", 4) == 0
C:/Users/Admin/Desktop/models 
 TRUE 
```

[...]

PR #1101, opened 2023-08-06
Fixing false file.access() errors on windows machines

marianklose:
## What?
I've provided a little bugfix for windows machines where `file.access()` falsely indicates readability issues. This can happen e.g., when working on mounted / mapped network drives. Please refer to issue #1100.

## Why?
The `new_build()` function checks via `file_readable()` (helper function for `file.access()`) if the directory is readable. `file.access()` however seems to be unreliable and throws errors even when the directory is readable. Thus, it is not justified to force a stop based on output from `file.access()`.

## How?
Minor change: changed the stop into a warning. Instead of forcing a stop, the user will be informed that the directory might be non-readable or non-existent.

## Testing
Tested the modified function on my local machine when working on a mapped network drive and now the model is correctly being read in plus throws a warning 

> Warning message:
> The file.access function indicates that project directory 'G:/Projekte/xxx/models' might not exist or is not readable. However, file.access is unreliable on Windows. Thus, existance and readability of the project directory should be ensured by the user.

 The change is not expected to affect any of the existing tests. devtools::check() passes without issues.

### DIFF
--- a/R/class_build.R
+++ b/R/class_build.R
@@ -69,7 +69,9 @@ new_build <- function(file=NULL, model, project, soloc=getwd(), code = NULL,
   env$soloc <- as.character(create_soloc(soloc,new_model,preclean))
   
   if(!file_readable(project)) {
-    stop("project directory '", project, "' must exist and be readable.",call.=FALSE) 
+    warning("The file.access function indicates that project directory '", project,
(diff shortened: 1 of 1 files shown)


kylebaron:
@marianklose  - 
Thanks for proposing the PR, but we won't be addressing the issue this way. Could you please close this and we'll be in touch to test the fix on your system.

Kyle

Issue #1103, opened 2023-08-14
Documented arguments not in \usage in documentation object ‘matlist’:


kylebaron:
This is new with rdevel

```
Version: 1.0.9
Check: Rd \usage sections
Result: NOTE
    Documented arguments not in \usage in documentation object ‘matlist’:
     ‘.x’
    
    Functions with \usage entries need to have the appropriate \alias
    entries, and all their arguments documented.
    The \usage entries must correspond to syntactically valid R code.
    See chapter ‘Writing R documentation files’ in the ‘Writing R
    Extensions’ manual.
```

This currently only shows up with `r-devel` and I can't install it to reproduce. I did test this on winbuilder and found that the change eliminates the NOTE.

kylebaron:
# After removing `.x` in documentation

```* using log directory 'd:/RCompile/CRANguest/R-devel/mrgsolve.Rcheck'
* using R Under development (unstable) (2023-08-14 r84947 ucrt)
* using platform: x86_64-w64-mingw32
* R was compiled by
    gcc.exe (GCC) 12.2.0
    GNU Fortran (GCC) 12.2.0
* running under: Windows Server 2022 x64 (build 20348)
* using session charset: UTF-8
* checking for file 'mrgsolve/DESCRIPTION' ... OK
* checking extension type ... Package
* this is package 'mrgsolve' version '1.1.0.9000'
* package encoding: UTF-8
* checking CRAN incoming feasibility ... NOTE
Maintainer: 'Kyle T Baron <kyleb@metrumrg.com>'

Version contains large components (1.1.0.9000)
* checking package namespace information ... OK
* checking package dependencies ... OK
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
* checking for hidden files and directories ... OK
* checking for portable file names ... OK
* checking whether package 'mrgsolve' can be installed ... OK
* used C++ compiler: 'g++.exe (GCC) 12.2.0'
* checking installed package size ... OK
* checking package directory ... OK
* checking for future file timestamps ... OK
* checking DESCRIPTION meta-information ... OK
* checking top-level files ... OK
* checking for left-over files ... OK
* checking index information ... OK
* checking package subdirectories ... OK
* checking R files for non-ASCII characters ... OK
* checking R files for syntax errors ... OK
* checking whether the package can be loaded ... OK
* checking whether the package can be loaded with stated dependencies ... OK
* checking whether the package can be unloaded cleanly ... OK
[...]

PR #1104, opened 2023-08-15
Fix NOTE from r-devel for .x in matlist documentation

kylebaron:
This is new with r-devel. I don't know exactly when, but let's reference `r84939` when I first noticed. I can't install r-devel right now and can only reproduce this via winbuilder. 

mrgsolve on cran: https://cran.r-project.org/web/checks/check_results_mrgsolve.html

Other packages are getting hit too: https://cran.r-project.org/web/checks/check_results_dplyr.html

See #1103 for before / after check results.

### DIFF
--- a/R/matlist.R
+++ b/R/matlist.R
@@ -244,7 +244,6 @@ setMethod("zero_re", "mrgmod", function(.x, ...) {
 
 #' Methods for working with matrix-list objects
 #'
-#' @param .x a matlist object
 #' @param x a matlist object
 #' @param ... passed through to other methods
 #'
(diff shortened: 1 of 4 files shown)


R/matlist.R
```diff
@@ -244,7 +244,6 @@ setMethod("zero_re", "mrgmod", function(.x, ...) {
 
 #' Methods for working with matrix-list objects
 #'
-#' @param .x a matlist object
```
kyleam, line 247: I didn't investigate the R devel change that started flagging this, but looks correct given that none of the (current) `*.matlist` methods have `.x` in their signatures.

kyleam (approved):

PR #1110, opened 2023-08-22
All records get a different draw from EPS

kylebaron:
Previous behavior was to assign a new `EPS` only when time was advancing in the data set within individual. Now, we assign a new `EPS` every record.

There was some linting that happened too ... not sure if this came from Rstudio or not. But let's leave it as is.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Data set records at the same time within individual will receive different 
   `EPS` draws; this is a change from previous behavior where records with the 
-  same time received the same value for `EPS` (#).
+  same time received the same value for `EPS` (#1110).
 
 # mrgsolve 1.1.1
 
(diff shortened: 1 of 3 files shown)


kyleam (approved):
-   The code change makes sense and looks straightforward (though I don't have a solid grasp of all the moving parts).

-   I confirmed the new test fails if I revert the code change.

-   I confirmed the internal cases that prompted this PR are resolved by this change (i.e. we no longer need to force any covariance matrices to be positive definite when calculating npde).

-   Reading through the internal discussion, it sounds like there is consensus about this change in behavior.

-   As far as I understand, this can be viewed as a pure bug fix. In that case, I think unconditionally changing the behavior along with a NEWS item is sufficient, as opposed to doing something more involved to retain the current behavior.

timwaterhouse (approved):
Looks good. Agree with @kyleam that the change along with the NEWS item is sufficient.

PR #1111, opened 2023-08-24
Release 1.2.0

kylebaron:
# mrgsolve 1.2.0

- Data set records at the same time within individual will receive different 
  `EPS` draws; this is a change from previous behavior where records with the 
  same time received the same value for `EPS` (#1110).


kyleam (approved):
I know this is stilled marked as a draft, but looks good to me whenever you're ready.

Issue #1117, opened 2023-09-27
mevent history doesn't get cleaned up properly


kylebaron:
User reported this problem with the mevent history: for the first ID, everything works fine; for the second ID, we don't get any of the events triggered from within the model because the event history clean up code is never reached in this case. 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '

$MAIN 
if(NEWIND <= 1) {
  int dosen = 0;
  double dose = 0;
}

$TABLE
if(EVID> 30 && dosen < 11) {
    if(dosen <= 1) dose = 100;
    if(dosen==2)   dose = 200;  
    if(dosen ==3)  dose = 300;
    if(dosen > 3)  dose = 400;
    if(dosen > 5)  dose =   0;
    mrg::evdata ev(TIME,1);
    ev.cmt = 1;
    ev.amt = dose;
    ev.now = true;
    self.mevector.push_back(ev);
    ++dosen;
    self.mevent(TIME + 3, 33);
}

$CMT A
$CAPTURE dosen dose
'

mod <- mcode("foo", code)
#> Building foo ...
#> done.
data <- expand.ev(amt = 0, evid = 33, time = 0, ID = 1:2)

mrgsim(mod, data = data, end = 33, delta = 0.1) %>% plot()
```

![](https://i.imgur.com/p8RvWKm.png)

<sup>Created on 2023-09-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

# Strategy to address

- Clear mevent history at ID reset
- Check `amt` in addition to `time`, `cmt`, `evid` 
- Maybe add a flag to not check the history

PR #1118, opened 2023-09-30
Fix mevent hx reset

kylebaron:
# Summary

We should clear the mtime event history with every subject but only if mtime is being used. Details in #1117


kyleam (approved):
`src/devtran.cpp` fix looks good to me. Verified regression test and reproducer from gh-1117 behave as expected before and after code change.

PR #1119, opened 2023-09-30
Refactor/mtimehx

kylebaron:
# Summary

The `evdata` object is a way to shuttle new dosing information from the model back to the mrgsolve simulation sequence. 

We are adding a `check_unique` attribute that allows users to skip checking the modeled event history. This lets you schedule identical events if needed. 

This PR also adds dose `amt` into the event history check; this allows users to schedule dosing events to the same compartment at the same time but with different dose amounts without tapping that `check_unique` attribute. 

At this point, tests are only focused on the `check_equal` feature.

Tests are getting added to a special C++ test space; this space will be added to the `drone` make target so tests will get run with push to drone.


kyleam (approved):
Looks good.  The new handling fits nicely into the current code.

As with gh-1125, conflict in `inst/maintenance/unit-cpp/test-cpp.R` is due to new tests here and new tests from gh-1131.  Confirmed that `make test-cpp` passes when all tests from both sides are included.

PR #1128, opened 2023-11-10
Fix/f lag

kylebaron:


kylebaron:
Closing this; the polymorphism was too complicated and probably unneeded.

Issue #1129, opened 2023-11-14
Incorrect F when F changes over the course of a lag time


kylebaron:
Current behavior: bioavailability is set from whatever is in play at the time of a dose. This creates a problem when lag times are being used and the dose is actually given at a time when bioavailability has changed since the apparent administration. 

In the example below, there is a lag time of 1 and both doses happen (apparently at time 0); so we should see 
`100 x 0.7 + 200 * 0.1 = 90` units in the dosing compartment at time 1. We don't see this because both doses see the 0.1 as the bioavailability. 

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

mod <- mread("1005", project = modlib())
#> Loading required namespace: xml2
#> Building 1005 ... done.

code <- '
$PARAM CL = 1, V = 10, KA = 1
$MAIN
ALAG_A = 1;
if(self.amt==100) F_A = 0.7; 
if(self.amt==200) F_A = 0.1;
$PKMODEL cmt = "A,B", depot = TRUE
'

mod <- mcode("foo", code)
#> Building foo ... done.

data <- data.frame(amt = c(100,200), evid = 1, time = 0, ID = 1, cmt = 1)

mrgsim(mod, data, recsort = 3, end = 1)
#> Model:  foo 
#> Dim:    4 x 4 
#> Time:   0 to 1 
#> ID:     1 
#>     ID time  A B
#> 1:   1    0  0 0
#> 2:   1    0  0 0
#> 3:   1    0  0 0
#> 4:   1    1 30 0
```

<sup>Created on 2023-11-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

One solution is bioavailability gets stored in the record object early on and is always read from that specific record object rather than this global bioavailability floating around.

PR #1130, opened 2023-11-14
Refactor bioavailability handling

kylebaron:
# Summary

This PR adds another member and methods to the `datarecord` class to track bioavailability (`Fn`). 

A while ago, we had two different types of records: observations and events and mrgsolve employed some polymorphism to get these to work together. 

https://github.com/metrumresearchgroup/mrgsolve/commit/a936b7e6c2d8a48aaff78b4daa17ff91c3d7c4be#diff-0e466cbf322abb6ca2747f70501b19721fced39bb696dfb1921ee40067bb5ece

I that setup, `Fn` was tracked in the event record object. 

Since then, we collapsed this down into a single record type and tracking all data in the object (including dose amounts for observation records). 

Part of that was to let `Fn` float around a bit, getting read from `$MAIN` when a record was getting processed and used in the various functions. 

See #1129 for an example of when this doesn't work. So this PR is moving `Fn` back into the object so that this value will follow the object. 

A bunch of tests were added too, including a key test for the problem that motivated this refactor. 
  
# Details
- `Fn` is now a member of `datarecord`
- Added `fn()` and `fn(double)`  - getter and setter
- Removed `Fn` that was getting passed into `sdnedule()` and `steady()`; this was never used
- Data members were re-ordered to put members with same type next to each other; this might be more space efficient
- One test in `test-z-alag-f.R` was specifically related to #1129; a bunch of other tests added to confirm behavior; a lot of this will also be confirmed in the NONMEM benchmarks

# To consider

- Because we had the polymorphism in the code _previously_, we were working with records as shared pointers. This isn't technically necessary now, but going to keep the code as is because (1) we might want to eventually go back to different record types and (2) there might be some advantage to using shared pointers for these records.


inst/include/datarecord.h
```diff
@@ -46,7 +46,7 @@ class datarecord {
              int pos_, double id_);   
   
   //! short event constructor
-  datarecord(short int cmt_, int evid_, double amt_, double time_, double rate_);
+  datarecord(short int cmt_, int evid_, double amt_, double time_, double rate_, double fn_);
```
kylebaron, line 49: This constructor is used when copying data record objects; adding `fn` as a reminder that this is a key item that needs to make it into the new object.

inst/include/datarecord.h
```diff
@@ -76,7 +76,7 @@ class datarecord {
   double rate(){return Rate;}
   void rate(double value) {Rate = value;}
   
-  double dur(double b);
+  double dur();
```
kylebaron, line 79: Infusion duration; no longer needs `Fn` to be passed.

inst/include/datarecord.h
```diff
@@ -115,21 +118,23 @@ class datarecord {
...
   void lagged() {Lagged = true;}
 
-  double Time; ///< record time
-  double Id; ///< record ID value
   int Pos; ///< record position number
   unsigned short int Evid; ///< record event ID
-  bool Output; ///< should this record be included in output?
-  bool Fromdata; ///< is this record from the original data set?
-  bool Lagged; ///< this record was added as result of ALAG
+  unsigned short int Ss; ///< record steady-state indicator
```
kylebaron, line 123: These got re-arranged to get all the doubles and bools next to each other.

src/devtran.cpp
```diff
@@ -518,14 +517,16 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
       if(this_rec->is_event()) {
         
         this_cmtn = this_rec->cmtn();
+
+        if(!this_rec->is_lagged()) {
```
kylebaron, line 521: This is key ... if this is a lagged dose (a phantom record) we _cannot_ change bioavailability here.

kyleam (approved):
I'm jumping ahead of the review request, but my understanding is that this is ready to go (with the already reviewed gh-1131 coming in on top).  The background and motivation for moving bioavailability into the data record makes sense to me, and the code change and extensive tests look nicely done.

PR #1131, opened 2023-11-18
Fix/infusion now

kylebaron:
# Summary 

Users are allowed to trigger discontinuities in the simulation (e.g. doses) from within the model. This is done through an `evdata` object which can be scheduled at some time in the future or the user can ask for it to happen `now`. 

There was a bug in this when an infusion was given by this now mechanism: there was no code to turn the infusion off. 

As I was addressing this issue, I realized that there are some other expected behaviors for these doses that weren't getting implemented when these `now` doses were triggered. 

# Details

## Workflow

- I was testing github actions and got a workflow file committed; intent is to remove that where ever it shows up

## `test-cpp.R`
- Unit tests for events triggered from within the model

## `devtran.cpp`

- This is where most of the action is
- `prob.rate_main()` can be called regardless of whether the record just came from the data or not; whenever the `rate` attribute is non-negative, this call will do nothing; added logic in `rate_main()` to make this clear
- Added some logic to do nothing with the modeled event when `evid` is `0`; this is an observation
- After creating the event (`new_ev`) we have to process it like other event records; get and check the bioavailability, call `rate_main()`, etc
- Added logic to turn off an infusion

## `odeproblem.cpp`

- Added code to return early if `rate` is not negative


kyleam (approved):
Thanks for the summary and details.  Some of `devtran.cpp` changes are a bit beyond my grasp but overall look good to me.  

I confirmed that `make test-cpp` passes on this PR and fails as expected when I revert the code changes.

Issue #1132, opened 2023-11-27
Format warnings


kylebaron:
>Dear maintainer,
> Please see the problems shown on
> <https://cran.r-project.org/web/checks/check_results_mrgsolve.html>.
> Please correct before 2023-12-11 to safely retain your package on CRAN.
> Best,
> -k

`NEWS.md` will be fixed under separate PR. 

```r
Version: 1.1.1
Check: whether package can be installed
Result: WARN 
  Found the following significant warnings:
    LSODA.cpp:208:32: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘long unsigned int’ [-Wformat=]
    LSODA.cpp:249:33: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:256:33: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:286:34: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:318:37: warning: format ‘%d’ expects argument of type ‘int’, but argument 2 has type ‘double’ [-Wformat=]
    LSODA.cpp:318:51: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘double’ [-Wformat=]
    LSODA.cpp:318:89: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘double’ [-Wformat=]
    LSODA.cpp:469:34: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:674:32: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:687:38: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
    LSODA.cpp:742:64: warning: format ‘%i’ expects argument of type ‘int’, but argument 2 has type ‘size_t’ {aka ‘long unsigned int’} [-Wformat=]
[...]

PR #1134, opened 2023-11-28
Match data type with format in sprintf

kylebaron:
See #1132 ... apparently new gcc setup is generating warnings because `sprintf()` isn't getting the right type. Most of the warnings were from `size_t`. There _should_ be a "length modifier" that you can use with unsigned integer (`%zu`), but gcc doesn't seem to like the `z` length modifier. Tried with this `%zi` as well, but gcc complains about `z`. 

I ended up casting `size_t` (unsigned long int) to `int` with C-style casts. I don't think this should be a problem ...  we should always be able to cast to int. 

On one of the `sprintf()` calls, I was passing `double` and trying to format with `%d` which is just wrong; updated that to use `%f`. 

**UPDATE**

- Casting to `int` seems like it worked although `unsigned long long` might be better.
-  It's possible that `%zu` will work now, but likely to work only on newest R releases. 
- Discussing with @kyleam, he suggested using `Rcpp::Rcerr` to write to `stderr`; this was easy to implement since the original `C++` code used `cerr` and seems to be the safest route to go at this point.

### DIFF
--- a/src/LSODA.cpp
+++ b/src/LSODA.cpp
@@ -205,7 +205,7 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
     if (neq <= 0)
     {
       //cerr << "[lsoda] neq = " << neq << " is less than 1." << endl;
-      REprintf("[lsoda] neq = %i is less than 1.\n", neq);
+      REprintf("[lsoda] neq = %zi is less than 1.\n", neq);
       terminate(istate);
       return;
     }
@@ -246,14 +246,14 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
       if (ml >= n)
       {
         //cerr << "[lsoda] ml = " << ml << " not between 1 and neq" << endl;
-        REprintf("[lsoda] ml = %i not between 1 and neq.\n", ml);
+        REprintf("[lsoda] ml = %zi not between 1 and neq.\n", ml);
         terminate(istate);
         return;
       }
       if (mu >= n)
       {
         //cerr << "[lsoda] mu = " << mu << " not between 1 and neq" << endl;
-        REprintf("[lsoda] mu = %i not between 1 and neq.\n", mu);
+        REprintf("[lsoda] mu = %zi not between 1 and neq.\n", mu);
         terminate(istate);
         return;
       }
@@ -283,7 +283,7 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
       if (ixpr > 1)
       {
         //cerr << "[lsoda] ixpr = " << ixpr << " is illegal" << endl;
-        REprintf("[lsoda] ixpr =%i is illegal.\n", ixpr);
+        REprintf("[lsoda] ixpr =%zi is illegal.\n", ixpr);
         terminate(istate);
         return;
       }
@@ -466,7 +466,7 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
         if (ewt[i] <= 0.)
         {
           //cerr << "[lsoda] ewt[" << i << "] = " << ewt[i] << " <= 0.\n" << endl;
-          REprintf("[lsoda] ewt[%i] = %g <= 0.\n", i, ewt[i]);
+          REprintf("[lsoda] ewt[%zi] = %g <= 0.\n", i, ewt[i]);
           terminate2(y, t);
           return;
         }
@@ -671,7 +671,7 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
           {
             // cerr << "[lsoda] " << mxstep << " steps taken before reaching tout"
             //      << endl;
-            REprintf("[lsoda] %i steps taken before reaching tout; consider increasing maxsteps.\n", 
+            REprintf("[lsoda] %zi steps taken before reaching tout; consider increasing maxsteps.\n", 
                      mxstep);
             *istate = -1;
             terminate2(y, t);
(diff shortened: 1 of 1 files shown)


src/LSODA.cpp
```diff
@@ -315,7 +315,7 @@ void LSODA::lsoda(LSODA_ODE_SYSTEM_TYPE f, const size_t neq, vector<double> &y,
         {
           // cerr << "[lsoda] tout = " << tout << " behind t = " << *t
           //      << ". integration direction is given by " << h0 << endl;
-          REprintf("[lsoda] tout = %d behind t = %d. integration direction is given by %d.\n", 
+          REprintf("[lsoda] tout = %.f behind t = %f. integration direction is given by %f.\n",
```
kyleam, line 318: Is the `.` in front of  the first `f` unintended?  (I think that will chop off everything after the decimal.)

kyleam:
> apparently new gcc setup is generating warnings because `sprintf()` isn't getting the right type.

Interesting.  I guess the warnings on the -devel run are likely surfaced by https://github.com/wch/r-source/commit/f867eddc38a8f1ae9ac74f5184cfa74f32dfce66

> Most of the warnings were from `size_t`. There _should_ be a "length modifier" that you can use with unsigned integer (`%zu`), but gcc doesn't seem to like the `z` length modifier. Tried with this `%zi` as well, but gcc complains about `z`.

My understanding is that GCC supports "%zu" for a size_t, and, testing with the gcc version on Metworx, that works fine.  Based on gh-1132, you were testing via <https://win-builder.r-project.org/win-builder>, so  perhaps the issue is with MinGW's GCC port not supporting it?

(edit: I see you just pushed a revert to the `%zu`, so perhaps not.  Can you give a link to the referenced discussion?)

> I ended up casting `size_t` (unsigned long int) to `int` with C-style casts. I don't think this should be a problem ... we should always be able to cast to int.

Hmm, probably not relevant given the switch back to `%zu`, but I guess the main issue is that size_t can be larger (e.g., unsigned long) than `int`?  Not sure if that'd be a concern for any of these spots.

```
#include <iostream>

using namespace std;

int main (int argc, char *argv[])
{
  size_t n;

  // n = 2147483647;
  n = 2147483648;
  printf("n = %zu\n", n);
  printf("n = %i\n", (int)n);
}
```

```
n = 2147483648
n = -2147483648
```

If `%zu` works everywhere mrgsolve is targeting, that's great.  Otherwise, the other thought I had is whether you could switch to using `Rcpp::Rcerr` to avoid worrying about printf portability?

kylebaron:
I don't think casting to `int` would be a problem ... these numbers that are getting printed can't get so big, but it does seem hacky. 

There was a discussion on R-pkg-devel yesterday where this was discussed: 
https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010131.html

Apparently he is patching GCC and this is making it into production on CRAN. 

I tried `%zu` on winbuilder and it came back with no warnings. 
https://win-builder.r-project.org/kJNwrO6x4EV6/00check.log

But I'm not sure what is the safest thing to do to avoid check warnings on re-submission.

src/LSODA.cpp
kylebaron: Thanks; didn't pick that up.

kylebaron:
@kyleam  - I just saw your suggestion for `Rcpp::Rcerr`;  that seems like a good way out of this. Let me try that out.

kyleam:
> I don't think casting to `int` would be a problem ... these numbers that are getting printed can't get so big, but it does seem hacky.

Okay.  And it seems like these are all informational messages, so there's not much harm if they somehow happened to exceed `int`.  (Given it's `size_t`, perhaps it should at least be `unsigned int` though.)

> There was a discussion on R-pkg-devel yesterday where this was discussed: https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010131.html

Thanks.  Oof, a lot of moving parts there to consider.

> I tried `%zu` on winbuilder and it came back with no warnings. https://win-builder.r-project.org/kJNwrO6x4EV6/00check.log

Oh, that's good.  When 54fa60d said "zu didn't work", I assumed it was for winbuilder.  What setup was that for?

For what it's worth, the `%zu` change (tested at f0638b09) got rid of the warnings on my end when running with

    * using R Under development (unstable) (2023-11-28 r85645)
    * using platform: x86_64-pc-linux-gnu
    * R was compiled by
        gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
        GNU Fortran (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

> But I'm not sure what is the safest thing to do to avoid check warnings on re-submission.

I guess for the CRAN Windows checks, you should feel confident if winbuilder works for you, based on what Tomas says [here](https://stat.ethz.ch/pipermail/r-package-devel/2023q4/010133.html):

    If you use the Winbuilder service maintained by Uwe Ligges, you will
    have the same setup as one of the platforms used for CRAN incoming checks.

But, right, I don't feel confident in saying "%zu" will work across the platform for all those checks.  If switching to Rcerr works for you, that's my leaning, but I have no objections if you feel better about either "%zu" or casting to int.

kylebaron:
>When https://github.com/metrumresearchgroup/mrgsolve/commit/54fa60ddb68eee1d32dc59db588071a36f9e3d66 said "zu didn't work", I assumed it was for winbuilder. What setup was that for?

I was only testing format options with winbuilder. I could be wrong, but I took that R-pkg-devel thread to mean that `%zu` would work only on the most recent winbuilder and I had initially tested when `%zu` wasn't supported but when I run now, it is supported. Could be wrong on that, though.

kyleam (approved):
I said:

> If switching to Rcerr works for you, that's my leaning, but I have no objections if you feel better about either "%zu" or casting to int.

... so I'll approve this now.  If you end up pushing the Rcerr changes, I'll of course provide a second set of eyes to look over the conversion.

kylebaron:
Thanks again for your review and suggestions, @kyleam; I did  switch the `REprintf` calls to `Rcpp::Rcerr`.  Will send this to `winbuilder` to see if anything gets flagged.

PR #1149, opened 2024-01-03
Fill out plugin for tools to implement events inside the model

kylebaron:
# Summary

## Housekeeping

-  `self` is an object with type`databox`; it is a struct that we use to shuttle information from the model back to `mrgsolve`; adding a `push()` method with signature `evdata` to push these event objects into a vector that will be processed by mrgsolve
- Move `evt` namespace out from under `mrgsolve`
- Add typedef `evt::ev` as an alias for `evdata`; I'm going to start shaping these to be more like the R version
- In `modelheader.h` and `model_include.R`, needed to switch around how the header file gets included; there will need to be some more refactoring of the include order; will take that up once this PR is merged; but for now, the tests show that `mrgsolve-evtools.h` is getting included

## New api

Signature rules

- `self` always comes first if it is in the signature
- `ev` comes first if `self` is not in the signature
- For doses, `<amt>` comes first, then `<compartment>`; `<rate>` comes after `<compartment>`

Two functions to implement doses now; they return nothing and there is no opportunity to interact with the object that implements the doses

- `void evt::bolus(self, <amt>, <compartment>)`: do a bolus dose right now 
- `void evt::infuse(self, <amt>, <compartment>, <rate>)`: do an infusion right now 

Two functions to construct dose objects; the object is returned to allow user to customize a bit

- `evt::ev bolus(<amt>, <compartment>)`: return an object for a bolus dose
- `evt::ev infuse(<amt>, <compartment>, <rate>)`: return an object for an infusion

Some functions to help working with returned objects

- `void retime(<ev>, <time>)`: set the time of the dose; this also sets `now` to `false`
- `void now(<ev>)`: set `now` to `true`
- `void push(self, <ev>)`: push an event object back to mrgsolve to process

Also added a `push` member function to `self`; this could be overloaded eventually to push different types of stuff back. This will always be available, regardless of whether or not you load the plugin.

There will definitely be more coming on this front. But I'm pretty sure on these items. I'd like to review this and merge and see how things go for what's next.

### DIFF
--- a/inst/maintenance/unit-cpp/test-evtools.R
+++ b/inst/maintenance/unit-cpp/test-evtools.R
@@ -103,5 +103,3 @@ test_that("evtools - give timed dose now", {
   b <- mrgsim(mod, param = list(mode = 1))
   expect_identical(as.data.frame(a), as.data.frame(b))
 })
-
-local_edition(2)
--- a/inst/maintenance/unit-cpp/test-evtools.R
+++ b/inst/maintenance/unit-cpp/test-evtools.R
@@ -42,6 +42,12 @@ if(mode==5 && givedose) {
   evt::bolus(self, dose, 1);
   evt::bolus(self, dose, 1);
 }
+if(mode==6 && givedose) {
+  evt::ev ev = evt::bolus(dose, 1);
+  ev.time = 100;
+  evt::now(ev);
+  evt::push(self, ev);
+}
 '
 
 mod <- mcode("test-evtools-model-1", code)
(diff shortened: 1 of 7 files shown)


inst/maintenance/unit-cpp/test-evtools.R
```diff
@@ -0,0 +1,107 @@
...
+  expect_equal(tmax, 0)
+})
+
+test_that("evtools - give timed dose now", {
+  a <- mrgsim(mod, param = list(mode = 6))
+  b <- mrgsim(mod, param = list(mode = 1))
+  expect_identical(as.data.frame(a), as.data.frame(b))
+})
+
+local_edition(2)
```
kyleam, line 107: Isn't this `local_edition` call unnecessary? The file-level edition set by the `local_edition` call on line 9 should be automatically reset.

kyleam (approved):
> New api

Nice.  Thanks for laying that out in detail.

> Move `evt` namespace out from under `mrgsolve`

Just to make sure I'm understanding correctly: that's a breaking change, correct?

It looks like this was only added in the 1.0.1 release (99f2aed3, gh-934), so is the expectation that the one function (`push`) didn't get much use in the wild?

inst/maintenance/unit-cpp/test-evtools.R
kylebaron: Ok; I wasn't sure about the scope for the editions.

kylebaron:
Thanks, @kyleam.  Yeah, this is a breaking change. We could easily leave the `mrgsolve::evt::push()` function in the code and not have it be breaking for now. 

I don't think there is currently a way to have this properly error right now (proper from R's perspective) ... in order to notify the user of the change. The other thing I could try is grepping this out of the model code and warning there. 

I'll put the code back in and think about how to handle this in the future.

kylebaron:
>so is the expectation that the one function (push) didn't get much use in the wild?
Yeah; that previous addition was just a small start to what I'm doing now; didn't get documented anywhere, except for the PR. 

I put that function back for now and will think of a way to pull it back someday.

PR #1150, opened 2024-01-05
Refactor default value for `check_unique`

kylebaron:
When an `evdata` object is constructed, lets set `check_unique` to `false` instead of the current default of `true`. This is what we want most of the time when users are constructing these objects themselves (for doses). 

For `mevent()` and `mtime()` we set `check_unique` to `true`; it was these functions that originally motivated this check  for duplicates and we can now control the check behavior when we calling these functions through `check_unique`. 

This is technically a breaking change but I think negative impact should be minimal; there just aren't many people using it. But we'll draw attention to this in the news.

### DIFF
--- a/inst/base/databox_cpp.h
+++ b/inst/base/databox_cpp.h
@@ -21,11 +21,13 @@
 
 void databox::mevent(double time, int evid) {
   mrgsolve::evdata ev(time,evid);
+  ev.check_unique = true;
   mevector.push_back(ev);
 }
 
 double databox::mtime(double time) {
   mrgsolve::evdata ev(time,2);
+  ev.check_unique = true;
   mevector.push_back(ev);
   return time;
 }
--- a/inst/base/mrgsolv.h
+++ b/inst/base/mrgsolv.h
@@ -54,7 +54,7 @@ struct evdata {
     amt = 0.0;
     rate = 0.0;
     now = false;
-    check_unique = true;
+    check_unique = false;
   } 
   double time; 
   int evid;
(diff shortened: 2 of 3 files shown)


kyleam:
With this default change, `make test-cpp` is hanging when I run it.  It gets stuck at the `mrgsim()` call here:

https://github.com/metrumresearchgroup/mrgsolve/blob/6ddefcb0fabbcd8a3b5f02db36bbf36403eb0c4b/inst/maintenance/unit-cpp/test-cpp.R#L90-L111

It looks like Drone is hitting into the same problem: https://github-drone.metrumrg.com/metrumresearchgroup/mrgsolve/1685/1/3

I can remove the hang (leading to all tests passing) with

```diff
diff --git a/inst/maintenance/unit-cpp/test-cpp.R b/inst/maintenance/unit-cpp/test-cpp.R
index 02e8c190..26a94102 100644
--- a/inst/maintenance/unit-cpp/test-cpp.R
+++ b/inst/maintenance/unit-cpp/test-cpp.R
@@ -97,6 +97,7 @@ if(TIME==0) {
   ev.amt = dose; 
   ev.now = n==1;
   ev.rate = r;
+  ev.check_unique = 1;
   self.mevector.push_back(ev);
 }
 if(TIME >= 3) {
```

I haven't dug deeper, but it seems likely that this switch is exposing a pre-existing bug?

kylebaron:
Thanks, @kyleam ; I ran in to this too when testing. I don't think it's a bug _per se_ in mrgsolve, but some careless code in the test that is no problematic with `check_unique` defaulting to `true`. 

I'll push a fix to the test to confirm this and maybe we can revisit the wisdom of this change :).

kyleam (approved):

kylebaron:
Thanks; I'll go forward with this for now and consider what mayhem might ensue.

Issue #1144, opened 2023-12-10
event 'until' usage bug/misunderstanding?


dpastoor:
I am trying to leverage until to design dosing regimens up til a fixed point in time where we're done observing. To use this example, given monthly dosing until 16 weeks I used the following - however my expectation was dosing would stop at 84 (or maybe 112 aka 16 weeks), instead I see that it continues out to 140 which is `until + ii` but that doesn't seem to align with the definition of 

> until | the expected maximum observation time for this regimen

```
> realize_addl(ev(amt = 300, time = 4*7, ii = 4*7, until = 16*7))
Events:
  time amt ii addl cmt evid
1   28 300  0    0   1    1
2   56 300  0    0   1    1
3   84 300  0    0   1    1
4  112 300  0    0   1    1
5  140 300  0    0   1    1
> 16*7
[1] 112
> 16*7+28
[1] 140

# what about a q2w dose:
> realize_addl(ev(amt = 300, time = 4*7, ii = 2*7, until = 16*7))
Events:
   time amt ii addl cmt evid
1    28 300  0    0   1    1
2    42 300  0    0   1    1
3    56 300  0    0   1    1
4    70 300  0    0   1    1
5    84 300  0    0   1    1
6    98 300  0    0   1    1
7   112 300  0    0   1    1
8   126 300  0    0   1    1
9   140 300  0    0   1    1
10  154 300  0    0   1    1
```

I'm not following how for the q4w regimen there is 1 extra dose past 112, but then if q2w there is 3 extra doses?

Would appreciate some clarity on if i'm misunderstanding `until`'s behavior?

kylebaron:
Sorry for late reply on this. Feels like a bug. I'm guessing there is some neglect of the fact that you are starting at `time = 4*7` or `time = 2*7` in the event. The results look ok if you start this at `time = 0`

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

realize_addl(ev(amt = 300, time = 4*7, ii = 4*7, until = 16*7))
#> Events:
#>   time amt ii addl cmt evid
#> 1   28 300  0    0   1    1
#> 2   56 300  0    0   1    1
#> 3   84 300  0    0   1    1
#> 4  112 300  0    0   1    1
#> 5  140 300  0    0   1    1
realize_addl(ev(amt = 300, time = 0, ii = 4*7, until = 16*7))
[...]

PR #1154, opened 2024-01-17
Fix until calculation when time not equal to 0

kylebaron:
@dpastoor reported bad behavior of `until` when constructing event objects #1144

The primary problem was that the logic inadvertently assumed the dose was given at time = 0. This PR fixes that logic. 

One legit question raised in the issue was when should dosing end when this works properly?  For example, `Q28d` dosing that runs `until` 168` hours. The dosing regimen ends at 168 hours; should there be a dose given at 168 hours or not?

As I was working on this issue, I was thinking there _should_ be a dose there. But it looks like up to this point, we have been assuming that there isn't a dose at that time. So I'm going to code this fix to respect that behavior even though I'm thinking about it a little different now. 

To be clear: the `until` argument is the expected last `observation` time covered by that dose sequence; so we keep you dosing to cover the `until` time. If a dose might be given at the same time as the last observation, `until` won't cover that. 

Using @dpastoor 's example, the dosing runs to 84 hours, not 112. But I agree it might be reasonable to run it to 112 too. 

``` r
library(mrgsolve)
#> 
#> 
#>     filter

realize_addl(ev(amt = 300, time = 4*7, ii = 4*7, until = 16*7-0.001))
#> Events:
#> 1   28 300  0    0   1    1
#> 2   56 300  0    0   1    1
#> 3   84 300  0    0   1    1
realize_addl(ev(amt = 300, time = 4*7, ii = 4*7, until = 16*7))
#> Events:
#> 1   28 300  0    0   1    1
#> 2   56 300  0    0   1    1
#> 3   84 300  0    0   1    1
realize_addl(ev(amt = 300, time = 4*7, ii = 4*7, until = 16*7+0.001))
#> Events:
#> 1   28 300  0    0   1    1
#> 2   56 300  0    0   1    1
#> 3   84 300  0    0   1    1
#> 4  112 300  0    0   1    1
```

<sup>Created on 2024-01-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

**EDIT**: this PR also modernizes the `ev()` help topic and adds some text and examples for what to expect for `until` behavior. There are some other modifications that I'd like to make, but this will require modifying some other help topics as well (e.g. documenting evid/EVID in data_set). I will do this as another PR.

### DIFF
--- a/R/events.R
+++ b/R/events.R
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 - 2022  Metrum Research Group
+# Copyright (C) 2013 - 2024  Metrum Research Group
 #
 # This file is part of mrgsolve.
 #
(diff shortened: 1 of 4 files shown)


dpastoor:
thanks kyle! I can also test kicking the tires on this if you'd like

tests/testthat/test-ev.R
```diff
@@ -260,3 +260,15 @@ test_that("until  issue-513", {
...
   expect_silent(mutate(ev(amt=100,ii=2,until=168),until=NULL,addl=5))
 })
+
+test_that("until with non-zero dose time gh-1144", {
+  # Every 4 weeks to 16 weeks; four total doses
+  e <- ev(amt = 100, ii = 4*7, until = 16*7)
+  expect_equal(e$addl, 3)
+  # Every 4 weeks to just under 16 weeks; three total doses
+  e <- ev(amt = 100, ii = 4*7, until = 16*7-0.01)
+  expect_equal(e$addl, 2)
```
kyleam, line 270: This assertion [fails][1]:

```
── 1. Failure ('test-ev.R:270:3'): until with non-zero dose time gh-1144 ───────
e$addl not equal to 2.
1/1 mismatches
[1] 3 - 2 == 1
```

Doesn't the actual value of 3 (four total doses) match the behavior in the PR description?

[1]: https://github-drone.metrumrg.com/metrumresearchgroup/mrgsolve/1706/1/3

kylebaron: Thanks @kyleam ; I thought that was passing but needs another look ... probably got mixed up when I realized the previous behavior wasn't what I as expecting this time around. I'll get it squared away and explain in the PR.

kyleam (approved):
Thanks for the updates.  LGTM

> As I was working on this issue, I was thinking there _should_ be a dose there. But it looks like up to this point, we have been assuming that there isn't a dose at that time. So I'm going to code this fix to respect that behavior even though I'm thinking about it a little different now.

I too would expect a dose there, but I agree with respecting the current behavior.

R/events.R
```diff
@@ -19,50 +19,48 @@
...
-#' @param object passed to show
+#' @param x a model object.
+#' @param time event time.
+#' @param amt dose amount.
+#' @param evid event ID.
+#' @param cmt compartment number or name.
+#' @param ID subject ID.
+#' @param replicate logical; if `TRUE`, events will be replicated for 
+#' each individual in `ID`.
+#' @param until the expected maximum **observation** time for this regimen;
```
dpastoor, line 33: thanks this is much clearer!

dpastoor (approved):
likewise LGTM - thank you!

thinking very much out loud and I am on the fence if this would even be that helpful....


one thought that this idea brings up, is a potential utility function `through` - at first I was thinking a separate param through, but that would clutter the api.

Thinking about a reference example of use of `fixed()` for use in stringr such that you can do like str_detect(strings, pattern) or str_detect(strings, fixed(pattern))

where `fixed` makes the comparison at the byte literal level instead of a regex.

ev(amt = 100, until = through(16*7)) where that would literally just be

through <- function(.x) {
   .x + 0.001
   }
   
   This would also setup through to be adaptable or have some heuristic based on time scales? Is that even a thing?? where in some situations incrementing by 0.0001 vs 0.01 vs 0.00000001 may actually meaningfully impact a choice?
   
  half of the value would be calling out that officially in usage where people may be more likely to read it - eg
  
  
  
image

is very clear vs having to find it in a nested example.

kylebaron:
Thanks for sketching that out, @dpastoor. This is definitely a "thing" in mrgsolve ... because we implement doses and ask for observations at well-ordered intervals, we frequently need to "decide" what to do when when these happen at the same time. Will keep thinking about this and see if any patterns emerge that are easy to grasp and clarify code. 

I think we're all on the same page that retaining current behavior is the right thing to do at this point. For future consideration, I do think this design decision is the right thing to do and what we want most of the time. Even though the `until` functionality seem like it should produce a dose at the `until` time, I see that time as the time that we want to either end the simulation or we want to start some other dosing regimen. For this example

```r
> ev(amt = 100, ii = 4*7, until = 16*7) %>% realize_addl()
Events:
  time amt ii addl cmt evid
1    0 100  0    0   1    1
2   28 100  0    0   1    1
3   56 100  0    0   1    1
4   84 100  0    0   1    1
> 16*7
[1] 112
```

We'd want to give the 100 mg dose `until` 16 weeks and at that point, start a new dose. In that case, we'd want to save `16*7` for the new dose. Maybe other ways to see it but I think this is the motivation behind the original decision. 

``` r
library(mrgsolve)
#> 
#> 
#>     filter
e1 <- ev(amt = 100, ii = 4*7, until = 16*7)
e2 <- ev(amt = 200, ii = 4*7, until =  32*7)
ev_seq(e1, e1)
#> Events:
#> 1    0 100 28    3   1    1
#> 2  112 100 28    3   1    1
```

<sup>Created on 2024-01-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #1156, opened 2024-01-23
Functionality in evt namespace to "auto" dose

kylebaron:
# Summary

As a user, I want to start and manage a dose regimen completely from within the model code. 

# Details

Implement a new class called `regimen` for now; this implements administration of doses in a regimen. This PR extends work done under #1149 

## Basic functionality

Methods:

- constructor (`regimen::regimen()`); the constructor doesn't get called with any arguments (we really can't have that), but several of the data members get initialized to sensible defaults via `reset()`
- `regimen::reset()` - sets sensible defaults
- `regimen::init()` - initializes the object; pass in `self` once and this gets held in the object; we also call `reset()` and set the `initialized` variable to `true`
  - We track the initialization call and don't do any doses unless the object has been initialized
- `regimen::execute()` - try to dose if it's time

__NOTE__ : we have to do some gymnastics to get around the fact that the `regimen` object needs to be declared  in `$GLOBAL`; this `init()` approach is what we commonly do. 

## Setters

- `regimen::amt(double)` - set the dose amount (data member `Amt`)
- `regimen::cmt(int)` - set the compartment number (data member `Cmt`)
- `regimen::ii(double)` - set the dose interval (data member `Ii`)
- `regimen::rate(double)` - make this non-zero to do an infusion (data member `Rate`)
- `regimen::until(double)` - when to stop dosing (data member `Until`)

## Getters

- `regimen::amt()` 
- `regimen::cmt()` 
- `regimen::ii()`
- `regimen::rate()`
- `regimen::until()`

## Other

There is a `regimen::flagnext()` function that tells `regimen` to add an _additional_ record that will force the simulation to stop at the _time of the next dose_. This is important if we want very sparse observation records in the data.

I added a `near()` function in the `evt` namespace for testing if we are at the right time to dose.

## Tests

I think I exercised all the functions at least once. The basic idea is that you can get the same result, either dosing through the data set or dosing inside the model. 

## Todo in the future

- Set up a system to throw an error through R; can't do this right now; but I'm thinking of passing something back through the `self` object

- More complexity in the regimen; for example, 21 days on, 7 days off; I'm not ready to do this now, but I _think_ we're still set up to do this sort of thing eventually

- We are getting more syntax like `object.member`; the pre-processor will sub these out with parameter names if there is overlap; we'll need some functionality to go through the code and find that syntax and add them to reserved words. 

# Example

``` r
library(mrgsolve)
#> 
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
code <- '
$PARAM CL = 2/10, V = 20, KA = 1, dose = 100, FLAG  = 0
interval = 4*7, last = 10*112

$PKMODEL cmt = "DEPOT,CENT", depot = TRUE

$PLUGIN evtools

$GLOBAL
evt::regimen reg;

$MAIN 
if(NEWIND <=1) {
  reg.init(self);
  reg.amt(dose);
  reg.ii(interval);
  reg.until(last);
  if(FLAG) reg.flagnext();
}

$ERROR
reg.execute();
capture CP = CENT/V;
'

mod <- mcode("foo", code, preclean = TRUE)
#> Building foo ...
#> done.

idata <- expand.idata(
  interval = c(4,8)*7, 
  dose = c(100, 300), 
  last = c(5,10,15)*112
)

mod %>% 
  update(end = 20*112) %>% 
  idata_set(idata) %>%
  carry_out(last,interval) %>% 
  mrgsim() %>% 
  plot(CP~time|factor(last)*factor(interval), scales = "same")
```

<sup>Created on 2024-01-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


R/Aaaa.R
```diff
@@ -93,7 +93,8 @@ Reserved <- c("ID", "amt", "cmt", "ii", "ss", "evid",
               "addl", "rate","time", Reserved_cvar,
               "AMT", "CMT", "II", "SS", "ADDL", "RATE", "THETA",
               paste0("pred_", c("CL", "VC", "V", "V2", "KA", "Q", "VP", "V3")),
-              "double", "int", "bool", "capture")
+              "double", "int", "bool", "capture", 
+              "until", "now")
```
kyleam, line 97: Is my understanding correct that 1) `until` is for the functions added in this PR and 2) `now` is for the function added in gh-1149?

inst/base/mrgsolve-evtools-regimen.h
```diff
@@ -0,0 +1,107 @@
...
+
+void regimen::execute() {
+  if(!initialized) return;
+  if(!evt::near(Self->time, dose_time)) return;
+  if(Self->time >= Until) return;
+  evt::infuse(*Self, Amt, Cmt, Rate);
+  prev_dose_time = dose_time;
+  dose_time = dose_time + Ii;
+  if(Flagnext) {
+    mrg::evdata ev(dose_time, 3333);
```
kyleam, line 95: This second parameter is `evid`, correct? And the actual value doesn't matter, and it's just chosen to be something well outside the normal set?

kyleam (approved):
I think I follow the high-level motivation.  Some of the mechanics and details are a bit beyond my grasp, but this looks good as far as I can tell, and the test setup/comparison is very nice.

R/Aaaa.R
kylebaron: Right; it took me a while to figure out there would be a clash here (the preprocessor picks up both `ev` and `now` in `ev.now`); I'm wondering about walking this back now that #1159 is in the mix; those should get picked up by this scan.

inst/base/mrgsolve-evtools-regimen.h
kylebaron: Right. I was going to let the user pass this in to be able to check for it later on, but I thought it was too much complexity at this point. 

There isn't currently a strict range for this, by my plan is to eventually do that. Right now it's just abnormal.

Issue #1143, opened 2023-12-10
Check code for `a.b` type syntax


kylebaron:
- Both lhs and rhs need to be excluded from possible parameter or compartment list

PR #1159, opened 2024-02-02
Check for clashes between model names and cpp dot syntax

kylebaron:
# Summary

I referred to this in #1156; original issue in #1143

# Example of what happens currently

Two examples below. In the first example, we have written some C++ code that clashes with parameter names; the model won't compile. 

Several of these names that we _might_ use are known and we have already reserved them. 

What I want to do in the PR is to find names of the first case and give them reserved status at the time of model load and prevent parameters or compartments or other model labels from using those names.

``` r
library(mrgsolve)
#> 
#> 
#>     filter

code <- '
$PARAM check_unique = 2
$MAIN 
mrg::evdata ev(0, 2);
ev.check_unique = true;
'

mod <- mcode("verify", code)
#> Building verify ...
#> error.
#> 
#> ---:: stderr ::---------------------------------------------
#> using C++ compiler: ‘Apple clang version 14.0.3 (clang-1403.0.22.14.1)’
#> using SDK: ‘MacOSX13.3.sdk’
#> 12:4: error: no member named '_THETA_' in 'mrgsolve::evdata'
#> ev.check_unique = true;
#> ~~ ^
#> ./verify-mread-header.h:38:22: note: expanded from macro 'check_unique'
#> #define check_unique _THETA_[0]
#>                      ^
#> 1 error generated.
#> make: *** [verify-mread-source.o] Error 1
#> 
#> ------------------------------------------------------------
#> Error: the model build step failed.

code <- '
$PARAM check_unique = 2, amt = 5
$MAIN 
mrg::evdata ev(0, 2);
ev.amt = 100;
'

mod <- mcode("verify2", code)
#> Error in validObject(.Object): invalid class "mrgmod" object: Reserved words in model names: amt
```

<sup>Created on 2024-02-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/class_mrgmod.R
+++ b/R/class_mrgmod.R
@@ -100,14 +100,18 @@ check_globals <- function(x,cmt) {
   return(invisible(NULL))
 }
 
-check_cpp_objects <- function(env, x) {
-  if(is.null(env[["cpp_objects"]])) return(NULL)
+check_cpp_dot <- function(env, x) {
+  # Vast majority of models will just return
+  if(is.null(env[["cpp_dot"]])) return(NULL)
+  cpp_dot <- setdiff(
+    env[["cpp_dot"]], 
+    x@envir[["MRGSOLVE_CPP_DOT_SKIP"]]
+  )
+  if(!length(cpp_dot)) return(NULL)
   mod_names <- names(x)
   mod_names <- mod_names[c("param", "init", "omega_labels", "sigma_labels")]
   mod_labels <- unlist(mod_names, use.names = FALSE)
-  cpp_objects <- env[["cpp_objects"]]
-  bad <- cpp_objects[cpp_objects %in% mod_labels]
-  if(!length(bad)) return(NULL)
+  bad <- cpp_dot[cpp_dot %in% mod_labels]
   bad <- setdiff(bad, Reserved)  
   if(!length(bad)) return(NULL)
   mod_names$sigma_labels <- unlist(mod_names$sigma_labels, use.names = FALSE)
@@ -115,8 +119,8 @@ check_cpp_objects <- function(env, x) {
   for(i in seq_along(bad)) {
     if(bad[i] %in% mod_names$param) bad[i] <- paste(bad[i], "(parameter)")
     if(bad[i] %in% mod_names$init) bad[i] <- paste(bad[i], "(compartment)")
-    if(bad[i] %in% mod_names$omega_labels) bad[i] <- paste(bad[i], "(eta name)")
-    if(bad[i] %in% mod_names$sigma_labels) bad[i] <- paste(bad[i], "(eps name)")
+    if(bad[i] %in% mod_names$omega_labels) bad[i] <- paste(bad[i], "(eta label)")
+    if(bad[i] %in% mod_names$sigma_labels) bad[i] <- paste(bad[i], "(eps label)")
   }
   if(length(bad) > 1) {
     msg <- "Reserved symbols cannot be used as model names:"
--- a/R/class_mrgmod.R
+++ b/R/class_mrgmod.R
@@ -117,10 +117,18 @@ check_cpp_dot <- function(env, x) {
   mod_names$sigma_labels <- unlist(mod_names$sigma_labels, use.names = FALSE)
   mod_names$omega_labels <- unlist(mod_names$omega_labels, use.names = FALSE)
   for(i in seq_along(bad)) {
-    if(bad[i] %in% mod_names$param) bad[i] <- paste(bad[i], "(parameter)")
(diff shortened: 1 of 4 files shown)


R/modspec.R
```diff
@@ -535,6 +535,24 @@ move_global2 <- function(spec, env, build) {
   spec
 }
 
+find_cpp_dot <- function(spec, env) {
+  to_check <- c("PREAMBLE", "MAIN", "PRED", "ODE", "TABLE", "GLOBAL")
+  x <- spec[names(spec) %in% to_check]
+  x <- unlist(x, use.names = FALSE)
+  # Narrow the search first; 10x speed up when searching for `pattern`
```
kyleam, line 542: nice

R/class_mrgmod.R
```diff
@@ -100,6 +100,47 @@ check_globals <- function(x,cmt) {
   return(invisible(NULL))
 }
 
+check_cpp_dot <- function(env, x) {
+  # Vast majority of models will just return
+  if(is.null(env[["cpp_dot"]])) return(NULL)
+  cpp_dot <- setdiff(
+    env[["cpp_dot"]], 
+    x@envir[["MRGSOLVE_CPP_DOT_SKIP"]]
```
kyleam, line 108: One of the tests specifies multiple `MRGSOLVE_CPP_DOT_SKIP ` values as comma-delimited.  Is this spot supposed to account for that?

tests/testthat/test-modspec.R
```diff
@@ -587,3 +587,64 @@ test_that("INPUT block", {
   expect_equal(tagdf$name, c("CL", "V2"))
   expect_equal(tagdf$tag, rep("input", 2))
 })
+
+test_that("Reserve names in cpp dot gh-1159", {
+  temp <- '$param {param}\n$main\n{main};\n$table\n{table}'
```
kyleam, line 592: Okay as is, but I think it'd be worth testing `init`, `omega_labels`, and `sigma_labels` too given that `check_cpp_dot` repeats some logic for those.

tests/testthat/test-modspec.R
```diff
@@ -587,3 +587,64 @@ test_that("INPUT block", {
...
+  temp <- '$param {param}\n$main\n{main};\n$env MRGSOLVE_CPP_DOT_SKIP="foo"'
+  
+  param <- "cl = 1, foo = 3, vc = 5"
+  main <- "double b = 5;\nfoo.bar = true;"
+  table <- "true;"
+  code <- glue::glue(temp)
+  
+  expect_is(mcode("cpp-dot-skip-1", code, compile = FALSE), "mrgmod")
+  
+  temp <- '$param {param}\n$main\n{main};\n$env MRGSOLVE_CPP_DOT_SKIP="foo,bar"'
```
kyleam, line 634: I'm confused by this `MRGSOLVE_CPP_DOT_SKIP="foo,bar"`.  As I note above, `check_cpp_dot` receives these as a single string ("foo,bar") and doesn't have any logic to split them.  And if that were to work, I'd expect the `bar` check on line 640 to not signal an error.  

If the `$env MRGSOLVE_CPP_DOT_SKIP="foo,bar"` were dropped, the assertions would make sense to me.  Perhaps that's leftover from a previous iteration?  (In that case, that still leaves the question of whether comma-delimited values should be handled.)

R/class_mrgmod.R
kylebaron: Thanks. Yes, this is where that should get handled, but does not.

tests/testthat/test-modspec.R
kylebaron: Thanks. Yeah, this doesn't make sense to me either now that look at it. I've refactored, including handling `MRGSOLVE_CPP_DOT_SKIP` properly.

tests/testthat/test-modspec.R
kylebaron: I refactored these tests to look at parameters, compartments, omega labels and sigma labels. The "problem" code is placed in preamble, main, ode, and table.

R/class_mrgmod.R
```diff
@@ -100,6 +100,49 @@ check_globals <- function(x,cmt) {
   return(invisible(NULL))
 }
 
+check_cpp_dot <- function(env, x) {
+  # Vast majority of models will just return
+  if(is.null(env[["cpp_dot"]])) return(NULL)
+  cpp_dot <- env[["cpp_dot"]]
```
kylebaron, line 106: refactored here

kylebaron:
@kyleam I refactored based on your review; hopefully this hits on everything and I did more good than harm.

kyleam:
@kylebaron Thanks for the updates!

kyleam (approved):

PR #1137, opened 2023-12-02
Release/1.3.0

kylebaron:
# mrgsolve 1.3.0

- The `evdata` object for modeled events now contains a `check_unique` member; 
  when set to `false`, the event will be processed without checking for a
  matching record in the modeled event log (#1119). 

- The `amt` attribute in `evdata` modeled event objects is now considered
  when looking for duplicate records in the modeled event log (#1119).

## Bugs Fixed

- Fixed a bug where multiple lagged doses given at the same time but with 
  different bioavailability were all given the bioavailability of the 
  last dosing record (#1129, #1130).

- Fixed a bug where modeled infusions given `now` were never turned
  off (#1131). 

- Fixed a bug where the `self` object (type: `databox`) could not be 
  passed into functions written into header files that were included
  through `$INCLUDE`; these header files are now included immediately
  preceding any user code written into `$GLOBAL`  (#1125, #1126). 

- Fixed a bug where modeled event log was not getting reset after 
  simulating the first individual; the resulted in events not getting
  executed in subsequent individuals when matching events were executed
  in the first individual; this affects modeled events only, not events
  coming from the data set (#1117, #1118).  

- Fixed a bug in detecting which compartments are receiving doses 
  (#1112, #1113).

PR #1161, opened 2024-02-03
Vignette

kylebaron:


seth127:
I assume this PR is what you're referencing in https://github.com/metrumresearchgroup/mrgsolve/pull/1137#issuecomment-1838685933 . I like this idea of including one simple one with the package and keeping the others as is. Let me know if we can do anything to help out with this.

Issue #1138, opened 2023-12-08
Considering simulating OMEGA regardless of etasrc


kylebaron:
The issue is that EPS isn't reproducible when simulating the same problem with etasrc set to `"OMEGA"` and `"data.all"`, for example. There isn't a bug here, it's just the random number generator isn't getting tapped in the same way to make EPS reproducible regardless of what we do with `ETA`. 

# Example of EPS that is not reproducible when switchin `etasrc`

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

data <- expand.ev(
  amt = 100, 
  ETA1 = 0.1, ETA2 = 0.2, ETA3 = 0.3
)

mod <- modlib("popex", capture = "EPS(1)")
#> Building popex ...
#> done.
mod <- smat(mod, matrix(1))

set.seed(112)
out1 <- mrgsim(mod, data)
set.seed(112)
out2 <- mrgsim(mod, data, etasrc = "data.all")

out1
#> Model:  popex 
#> Dim:    482 x 10 
#> Time:   0 to 240 
#> ID:     1 
#>     ID time    GUT  CENT    CL     V   ECL  IPRED      DV   EPS_1
#> 1:   1  0.0   0.00  0.00 1.458 22.52 0.377 0.0000  0.0000 -0.9965
#> 2:   1  0.0 100.00  0.00 1.458 22.52 0.377 0.0000  0.0000 -0.2696
#> 3:   1  0.5  81.24 18.45 1.458 22.52 0.377 0.8193  0.2427 -1.2165
#> 4:   1  1.0  66.00 32.85 1.458 22.52 0.377 1.4588  0.7215 -0.7040
#> 5:   1  1.5  53.61 43.99 1.458 22.52 0.377 1.9531  4.1797  0.7608
#> 6:   1  2.0  43.55 52.48 1.458 22.52 0.377 2.3301  2.9877  0.2486
#> 7:   1  2.5  35.38 58.84 1.458 22.52 0.377 2.6128 19.0116  1.9846
#> 8:   1  3.0  28.74 63.50 1.458 22.52 0.377 2.8194 17.3175  1.8152
out2
#> Model:  popex 
#> Dim:    482 x 10 
#> Time:   0 to 240 
#> ID:     1 
#>     ID time    GUT  CENT    CL     V ECL  IPRED     DV   EPS_1
[...]

PR #1163, opened 2024-02-04
Always simulate from OMEGA so EPS is reproducible

kylebaron:
See examples in #1138

The PR changes behavior so that we always simulate from OMEGA, even when requesting ETA to come from an input data set. This will make EPS reproducible between a run that uses simulated ETA and a run that pulls ETA from the data set. 

This change will change certain simulations involving RUV and ETAs pulled from the input data. I don't expect many people to be affected at this point in the game.


kyleam (approved):
Motivation makes sense to me, code change lines up with what I'd expect based on quck-and-dirty patch in metrumresearchgroup/bbr.bayes#118, and the test looks good.

Thanks @kylebaron.

PR #1164, opened 2024-02-04
Drop specialized comment character in articles

kylebaron:
Merging this out-of-package documentation fix; code base is not affected.


kylebaron:
Screenshot 2024-02-03 at 21 40 20

PR #1187, opened 2024-04-19
Intelligent insertion of records for lagged doses and infusion stop events

kylebaron:
# Summary 

Every time we start an infusion, we need to figure out when it ends and insert a record for that to happen. This is complicated by the fact that we aren't guaranteed to know the infusion duration for any given dose until we actually give that dose. 

Current behavior is to push this dose record on the back of the `vector` / `deque` and then re-sort the records. This is fine when the data set is set up to dose in bulk, but really slows down when doses are explicit in the data set and have to be processed one at a time. 

By controlled or intelligent insert, I mean we know when the infusion end record happens, so we iterate through the next records until we find the correct insertion point and insert there. With this approach, there is no need to sort. This, along with the `deque` speeds things up a lot in specific circumstances. 

It's a similar story with doses involving lag time: once we know the lag time, we have to insert a record a the time of the original dose plus the lag time. So it's the same problem.

### DIFF
--- a/src/datarecord.cpp
+++ b/src/datarecord.cpp
@@ -582,3 +582,19 @@ void datarecord::schedule(std::vector<rec_ptr>& thisi, double maxtime,
     thisi.push_back(evon);
   }
 }  
+
+void insert_record(std::vector<rec_ptr>& thisi, const int start, rec_ptr& rec) {
+  double time = rec->time();
+  int i = start;
+  for(i = start; i < thisi.size(); ++i) {
+    if(thisi.at(i)->time() >= time) {
+      break;  
+    }
+  }
+  // Rcpp::Rcout << " start " << start << std::endl;
+  // Rcpp::Rcout << " i " << i << std::endl;
+  // Rcpp::Rcout << " size " << thisi.size() << std::endl;
+  
+  reclist::iterator alagit = thisi.begin() + i;
+  thisi.insert(alagit, rec);
+}
--- a/src/datarecord.cpp
+++ b/src/datarecord.cpp
@@ -584,9 +584,12 @@ void datarecord::schedule(std::vector<rec_ptr>& thisi, double maxtime,
 }  
 
 /* 
- * Inserts a record at the first opportunity based on time. Boolean flag
- * indicates if the record should get inserted before or after all other 
- * records at the time of that event. 
+ * Inserts a record at the earliest or latest opportunity based on time. 
+ * Boolean flag indicates if the record should get inserted before or after all 
+ * other records at the time of that event. 
+ * 
+ * This is currently being used to insert records for lag times as well as
+ * infusion end. 
  * 
  */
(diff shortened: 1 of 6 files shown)


src/datarecord.cpp
```diff
@@ -582,3 +582,32 @@ void datarecord::schedule(std::vector<rec_ptr>& thisi, double maxtime,
...
+/* 
+ * Inserts a record at the earliest or latest opportunity based on time. 
+ * Boolean flag indicates if the record should get inserted before or after all 
+ * other records at the time of that event. 
+ * 
+ * This is currently being used to insert records for lag times as well as
+ * infusion end. 
+ * 
+ */
+void insert_record(reclist& thisi, const int start, rec_ptr& rec,
```
kylebaron, line 595: New with this PR.

src/devtran.cpp
```diff
@@ -559,12 +560,13 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
...
-            advance(alagit,1);
-            a[i].insert(alagit,newev);
+            //reclist::iterator alagit = a[i].begin()+j;
+            //advance(alagit,1);
+            //a[i].insert(alagit,newev);
+            insert_record(a[i], j, newev, put_ev_first);
             newev->schedule(a[i], maxtime, put_ev_first, NN, prob.alag(this_cmtn));
             this_rec->unarm();
-            sort_recs = true;
+            sort_recs = newev->needs_sorting();
```
kylebaron, line 569: Needs sorting only if there were additional doses given in `schedule()`. In this case, the records are pushed and we have to sort. I thought this wasn't a huge problem because we usually put in a bunch of doses at once and only have to do it limited number of times, rather every dose. 

Might look at using insert dose here too eventually.

src/devtran.cpp
```diff
@@ -585,9 +587,12 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
...
-          }
-          a[i].push_back(evoff);
-          sort_recs = true;
+          } 
+          // TODO: drop when ready
+          //a[i].push_back(evoff);
+          //sort_recs = true;
+          // Infusion off always happens first
+          insert_record(a[i], j, evoff, true);
         }
```
kylebaron, line 596: This inserts the infusion end record.

src/datarecord.cpp
kylebaron: The if/else code is almost identical, just a matter if we have to exceed the target time or not (in case there is already an event at the target time).

kyleam (approved):
Thanks for the high-level motivation and inline comments.  Makes sense to me.

Changes for this PR ([this diff](https://github.com/metrumresearchgroup/mrgsolve/compare/3d5d70189b61345478f154de30000c9288b886e8...573e9c704d93edcb0c4f31f8e9666860d1e298fd) plus tip commit removing comments) look good to me.

Issue #1177, opened 2024-03-12
Export or print updated model code


RichardHooijmaijers:
I have a case where I want to update a model with for instance new parameter values. I also want to print or save this updated model code, but struggling to do so. For instance:

```
library(mrgsolve)
# Load a model
mod  <- house()
# Update a parameter 
mod2 <- param(mod,CL=2)

# This will all print the original model code
mod2$code 
mrgsolve:::code(mod2)  
see(mod2, raw=TRUE)
# This line does not work, also the example in the help file fails
# mrgsolve:::render(mod2)

``` 

Is there a way to print the entire model code where the new parameter value (CL=2) is included?
Thanks!

jimm3134:
workaround - 
use mrgsolve:::code(mod2) to print to the console, then select that code with your mouse and copy it to a document for printing.

RichardHooijmaijers:
Hi @jimm3134 thanks for your reply. I used `mrgsolve:::code(mod2)` in my example. And although this prints the model, it is the original model. I would like to know if there is a way which shows the updated model (e.g. where CL is set to 2)

kylebaron:
Hi @RichardHooijmaijers  - I'm sorry for the delay responding to your question. There isn't currently a way to write the updated object back to the model specification format. The best you can do is save the object to, say, `.rds` file and then read it back in when needed. `mrgsolve:::code()` (or `mod@code`) always goes back to the original model code. I am working on a feature to write the different model components out in `.json` format to be read back in; that would also preserve updated components of the model. But it would give you a version of the model code that is transportable and readable. 

Kyle

RichardHooijmaijers:
Hi @kylebaron thanks for clarifying! I really like the idea of using json format including options for updated components. Will keep an eye out for this.

kylebaron:
@RichardHooijmaijers this is now released to CRAN:
https://cran.r-project.org/package=mrgsolve

I'm expecting we'll need some refinement of some sort; so please open an issue if you see anything that could be improved.

PR #1190, opened 2024-05-06
mrgsolve model transport format

kylebaron:
# Summary

Background: I've been wondering about something like this or a while so it would be easier to share models that were dependent on NONMEM outputs. Then Richard asked about this in #1177 and we got motivated to really try to push this out. 

Convert an mrgsolve model object to a "transport" file format. I was originally aiming to do this in json format, but tried yaml too and yaml seemed (1) much easier to read (not sure if there is anything else I can do to improve rendering of the json) and (2) more reliable in rendering the object. So, I'm providing methods for either format. 

The _main_ feature here is that it frees you from requiring NONMEM output when reading in the model file. But you can also see _almost_ everything in the model object, including, for example, parameters or other settings that have been updated. And you _could_ read in this transport format, programmatically make changes to a parameter or setting, then write back. You can now and unlikely you'll ever be able to re-write actual model code (not what this is intended for).

- Write: `mwrite_yaml()` and `mwrite_json()`
- Read: `mread_yaml()` and `mread_json()`; these convert the yaml (or json) to mrgsolve model code, compile, and load
- Also providing functions to just convert yaml (or json) to mrgsolve model code (`yaml_to_cpp()` and `json_to_cpp()`)

__EDIT__: removed all json functionality on code review; may revisit at another time _if_ there is some need that can be met with json and not yaml. 

The transport format tracks
- `source` - the transport code is rejected if this is not `mrgsolve::mwrite`
- `mrgsolve` version number (not currently used)
- transport formate `version` (not currently used)
 
## Digits

Writers for both yaml and json accept a `digits` argument; this is defaulting to 8 right now. 

## PKMODEL

We do some minor gymnastics with the `PKMODEL` block. This is because you can declare compartments in `$PKMODEL` and we need to keep `PKMODEL` in the model spec, but need to remove the compartment declaration
since we are taking compartments from the model `init()`. 

## SET

Also some work to coordinate the `$SET` block. There are a bunch of slots in the model object that we'd like to carry into the transport file and then bring back when we load the model from transport. When calling `mread_yaml()`, we do that update from a block of settings in the transport file (under `update`). But I also wanted to be able to write all that into the file. So, on write, we parse out `$SET` and track that in the transport file. When requested, we can append to that model code with these `update` items. 

## CAPTURE
Late decision to explicitly handle capture items in the list / yaml.  This means `$CAPTURE` isn't going to appear in the general `code` item, but rather in a `capture` item, along the lines of `$PARAM` and `$CMT`.  As far as I can tell, this means that no annotations will appear in the model after it passes through yaml / list. 

# Example

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(mrgsolve)
#> 
#> 
#>     filter
```

Load a regular mrgsolve model

``` r
mod <- modlib("1005")
#> Loading required namespace: xml2
#> Building 1005 ... done.
```

# Write

We can write it out in `yaml` format

``` r
y <- mwrite_yaml(mod, file = "1005.yaml")
```

The yaml is easier to read

``` r
cat(readLines("1005.yaml"), sep = "\n")
#> source: mrgsolve::mwrite
#> mrgsolve: 1.4.1.9000
#> format: yaml
#> version: 1.0
#> model: '1005'
#> prob:
#> - 1005 phase1 2 CMT like 1004 but diff. initial on V3
#> - Run
#> - file.show(system.file("nonmem", "1005", "1005.ctl", package = "mrgsolve"))
#> - for equivalent NONMEM control stream.
#> param:
#>   SEX: 0.0
#>   WT: 70.0
#>   THETA1: 9.50788555
#>   THETA2: 22.79098949
#>   THETA3: 0.07143366
#>   THETA4: 3.47450589
#>   THETA5: 113.27671224
#>   THETA6: 1.02435433
#>   THETA7: 1.19211818
#> init:
#>   GUT: 0.0
#>   CENT: 0.0
#>   PERIPH: 0.0
#> omega:
#>   data:
#>     '...':
#>     - 0.21387884
#>     - 0.1207702
#>     - 0.09451047
#>     - -0.01162777
#>     - -0.03720637
#>     - 0.04656315
#>   labels:
#>     '1':
#>     - '.'
#>     - '.'
#>     - '.'
#>   names: '...'
#> sigma:
#>   data:
#>     '...':
#>     - 0.04917071
#>     - 0.0
#>     - 0.20176885
#>   labels:
#>     '1':
#>     - '.'
#>     - '.'
#>   names: '...'
#> env: []
#> plugin: base
#> capture:
#> - CL
#> - Q
#> - V2
#> - V3
#> - KA
#> - ETA_1
#> - ETA_2
#> - ETA_3
#> - IPRED
#> update:
#>   start: 0.0
#>   end: 24.0
#>   delta: 1.0
#>   add: []
#>   atol: 1.0e-08
#>   rtol: 1.0e-08
#>   ss_atol: 1.0e-08
#>   ss_rtol: 1.0e-08
#>   maxsteps: 20000.0
#>   hmax: 0.0
#>   hmin: 0.0
#>   mxhnil: 2.0
#>   ixpr: 0.0
#>   mindt: 2.22044605e-15
#>   digits: -1.0
#>   tscale: 1.0
#>   outvars:
#>   - GUT
#>   - CENT
#>   - PERIPH
#>   - CL
#>   - Q
#>   - V2
#>   - V3
#>   - KA
#>   - ETA_1
#>   - ETA_2
#>   - ETA_3
#>   - IPRED
#> set: []
#> code:
#> - $PKMODEL
#> - depot = TRUE
#> - ncmt = 2
#> - ' '
#> - $PK
#> - double CL = THETA(1)*exp(ETA(1)) * pow(THETA(6),SEX) * pow(WT/70.0,THETA(7));
#> - double V2 = THETA(2)*exp(ETA(2));
#> - double KA = THETA(3)*exp(ETA(3));
#> - double Q  = THETA(4);
#> - double V3 = THETA(5);
#> - double S2 = V2;
#> - ' '
#> - $ERROR
#> - double F = CENT/S2;
#> - double Y = F*(1+EPS(1)) + EPS(2);
#> - double IPRED = F;
#> - ' '
#> - $CAPTURE
#> - CL Q V2 V3 KA ETA(1) ETA(2) ETA(3) IPRED
#> - ' '
```

# Read

We can read the model back from yaml format

``` r
mod <- mread_yaml("1005.yaml")
#> Building 1005_yaml_mod ... done.

mod
#> 
#> 
#> --------------  source: 1005.yaml.mod  --------------
#> 
#>   project: /private/var/fol.../T/RtmpnYp7PE
#>   shared object: 1005.yaml.mod-so-aef1758e6b06 
#> 
#>   time:          start: 0 end: 24 delta: 1
#>                  add: <none>
#> 
#>   compartments:  GUT CENT PERIPH [3]
#>   parameters:    SEX WT THETA1 THETA2 THETA3 THETA4
#>                  THETA5 THETA6 THETA7 [9]
#>   captures:      CL Q V2 V3 KA ETA_1 ETA_2 ETA_3 IPRED [9]
#>   omega:         3x3 
#>   sigma:         2x2 
#> 
#>   solver:        atol: 1e-08 rtol: 1e-08 maxsteps: 20k
#> ------------------------------------------------------

mrgsim(mod, ev(amt = 100, ID = 1:10)) %>% plot()
```

We can also convert from the yaml file to a mrgsolve cpp file

``` r
x <- yaml_to_cpp("1005.yaml", project = getwd())

cat(readLines("1005.yaml.mod"), sep = "\n")
#> $PROB
#> 1005 phase1 2 CMT like 1004 but diff. initial on V3
#> Run
#> file.show(system.file("nonmem", "1005", "1005.ctl", package = "mrgsolve"))
#> for equivalent NONMEM control stream.
#> 
#> $PARAM
#> SEX = 0
#> WT = 70
#> THETA1 = 9.50788555
#> THETA2 = 22.79098949
#> THETA3 = 0.07143366
#> THETA4 = 3.47450589
#> THETA5 = 113.27671224
#> THETA6 = 1.02435433
#> THETA7 = 1.19211818
#> 
#> $INIT
#> GUT = 0
#> CENT = 0
#> PERIPH = 0
#> 
#> $OMEGA
#> @block
#> @labels . . .
#> 0.21387884
#> 0.1207702
#> 0.09451047
#> -0.01162777
#> -0.03720637
#> 0.04656315
#> 
#> $SIGMA
#> @block
#> @labels . .
#> 0.04917071
#> 0
#> 0.20176885
#> 
#> $PKMODEL
#> depot = TRUE
#> ncmt = 2
#>  
#> $PK
#> double CL = THETA(1)*exp(ETA(1)) * pow(THETA(6),SEX) * pow(WT/70.0,THETA(7));
#> double V2 = THETA(2)*exp(ETA(2));
#> double KA = THETA(3)*exp(ETA(3));
#> double Q  = THETA(4);
#> double V3 = THETA(5);
#> double S2 = V2;
#>  
#> $ERROR
#> double F = CENT/S2;
#> double Y = F*(1+EPS(1)) + EPS(2);
#> double IPRED = F;
#>  
#> $CAPTURE
#> CL Q V2 V3 KA ETA(1) ETA(2) ETA(3) IPRED
#>  
#> $SET
#> start = 0
#> end = 24
#> delta = 1
#> add = numeric(0)
#> atol = 1e-08
#> rtol = 1e-08
#> ss_atol = 1e-08
#> ss_rtol = 1e-08
#> maxsteps = 20000
#> hmax = 0
#> hmin = 0
#> mxhnil = 2
#> ixpr = 0
#> mindt = 2.22044605e-15
#> digits = -1
#> tscale = 1
#> outvars = c("GUT", "CENT", "PERIPH", "CL", "Q", "V2", "V3", "KA", "ETA_1", "ETA_2", "ETA_3", "IPRED")

mread("1005.yaml.mod")
#> Building 1005_yaml_mod ... done.
#> 
#> 
#> --------------  source: 1005.yaml.mod  --------------
#> 
#>   project: /private/var/fol...0-godly-vixen
#>   shared object: 1005.yaml.mod-so-aef16fbe40b6 
#> 
#>   time:          start: 0 end: 24 delta: 1
#>                  add: <none>
#> 
#>   compartments:  GUT CENT PERIPH [3]
#>   parameters:    SEX WT THETA1 THETA2 THETA3 THETA4
#>                  THETA5 THETA6 THETA7 [9]
#>   captures:      CL Q V2 V3 KA ETA_1 ETA_2 ETA_3 IPRED [9]
#>   omega:         3x3 
#>   sigma:         2x2 
#> 
#>   solver:        atol: 1e-08 rtol: 1e-08 maxsteps: 20k
#> ------------------------------------------------------
```

<sup>Created on 2024-07-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


kyleam:
This looks very neat.  Thanks for the detailed motivation and overview.

I've digested this only at a high level and won't get to a full review today, but I'm curious about this part:

> I was originally aiming to do this in json format, but tried yaml too and yaml seemed (1) much easier to read (not sure if there is anything else I can do to improve rendering of the json) and (2) more reliable in rendering the object. So, I'm providing methods for either format.

 * For the "more reliable" rendering, what cases in particular tripped up the json variant?

 * Why offer both variants instead of just yaml?  
    
    Are there scenarios where the json variant is preferable to the yaml variant?  My thinking would be that offering the json variant isn't worth the extra code and documentation unless it provides some value over the yaml variant (beyond subjective format preference).  And given that your *2* indicates that it has known limitations, it seems likely that providing the json variant does more harm than good.

kylebaron:
Thanks, @kyleam. 


> For the "more reliable" rendering, what cases in particular tripped up the json variant?

Maybe "reliable" isn't the right word. It was subtle differences in how R objects would get rendered in json or reconstituted when going back to R. I was surprised at how many different ways stuff could get rendered ... a little bit art. But now that seems very naive. Specifically,  I think there might have been some unexpected way it handled empty objects (like lists or vectors).  Maybe something with the attributes. It was all pretty minor and I was able to write workarounds to handle it and I think I've addressed, so I'm not totally sure how much of the clean up code is only devoted to getting json to work. 

> Why offer both variants instead of just yaml?

Good question; I don't have a great answer. I was going into this thinking json and it seem like the format all the cool kids are using. I wouldn't be against going with just yaml, but maybe a little uncertain based on what I wrote above about reliability that I'll notice something funny about the rendering down the road and wish I would have gone with the "other" one. I did try to minimize the amount of code duplication (doing all the work outside of actually reading and writing and rendering with common functions), but yeah there are more functions to call and maintain. 

I'd be happy to go with just yaml for this to minimize future maintenance and review if that's what you'd advise; it should be minor effort and I could do that prior to reviewing all the code to reduce the code review burden. 

KTB

kyleam:
> Maybe "reliable" isn't the right word. [...]

Okay, thanks.

> I wouldn't be against going with just yaml, but maybe a little uncertain based on what I wrote above about reliability that I'll notice something funny about the rendering down the road and wish I would have gone with the "other" one.

I see.

> I did try to minimize the amount of code duplication (doing all the work outside of actually reading and writing and rendering with common functions), but yeah there are more functions to call and maintain.

True, the shared functionality is nicely separated.  (I had considered mentioning that in my original comment.)

I suppose I could see that as a reason to not worry about the "wish I would have gone with the other one" thing.  The underlying structure can stay the same even if you drop the json stuff, so it's easy enough to add back if needed.

> I'd be happy to go with just yaml for this to minimize future maintenance and review if that's what you'd advise

No, my question wasn't considering code review.  I've no problems reviewing it as is, and, as you mention, the meat is elsewhere.

If I had to choose, I'd lean toward not adding another public-facing piece unless there is a concrete reason (in particular something that makes it not redundant).  But I don't have any strong objections to exposing both the json and yaml functionality, and I'll happily leave that choice to you.

kylebaron:
@kyleam   - as best as I can tell, the json stuff is gone. I updated one error message and a test. And cleaned out the documentation too.

kylebaron:
Update to handle `$CAPTURE`: 

Currently, I'm injecting a table of model annotations into the `$PROBLEM` block to save that information b/c it's getting clobbered as we recast $PARAM and `$CMT` blocks. 

`$CAPTURE` can also be annotated, so this will cause a problem - model annotations will _keep_ getting appended to the text in `$PROBLEM` if we keep (re)saving to yaml and read in again; it will keep getting duplicated (at least the `$CAPTURE` stuff will). 

Rather than checking if model annotations are already inserted into `$PROBLEM`, I'm going to explicitly handle `$CAPTURE` information in the list / yaml structure. These will get written to yaml and then back into the mrgsolve control stream, getting sanitized of the annotations along the way. 

Now, all potentially annotated blocks are getting clobbered and there will be no more annotations the second time we read the model (except for the annotations that we already cached in `$PROBLEM`). 

So we can say this model transport process

1. Saves annotations in `$PROBLEM`
2. Drops all annotations _in the code_ (so stuff from 1. won't be duplicated)
3. Has consistent handling for `$PARAM`, `$CMT`, and `$CAPTURE` ... they show up explicitly in the yaml / list, not in the chunk of code that is taken as-is (not parsed or explicitly handled in the list / yaml).

R/mwrite.R
```diff
@@ -0,0 +1,341 @@
...
+  l$version <- 1
+  l$model <- x@model
+  # Problem
+  l$prob <- ""
+  # Critical items
+  l$param <- as.list(param(x))
+  l$init <- as.list(init(x))
+  l$capture <- character(0)
+  if(length(x@capture)) {
+    capture <- names(x@capture)
```
kylebaron, line 36: Some funny handling here ... 

$CAPTURE items can be renamed as

```
$CAPTURE new = old
```

what we store in `x@capture` is the new name, with the old names in the `names` attribute.

kylebaron: It looks like this

```r
> mcode("test", "$PARAM old = 2\n$CAPTURE new = old", compile = FALSE)@capture
  old 
"new" 
```

R/mwrite.R
```diff
@@ -0,0 +1,341 @@
...
+    }
+  }
+  
+  if("PROB" %in% names(code)) {
+    l$prob <- code$PROB
+    code$PROB <- NULL
+  }
+  
+  clob <- c("PARAM", "INPUT", "THETA", "CMT", "INIT", "OMEGA", "SIGMA", 
+            "NMEXT", "NMXML", "VCMT", "SET", "CAPTURE")
```
kylebaron, line 103: Clobbering `$CAPTURE` now b/c it is explicitly addressed.

R/mwrite.R
```diff
@@ -0,0 +1,341 @@
...
+  for(i in seq_along(x$sigma$data)) {
+    header <- "@block"
+    if(any(x$sigma$labels[[i]] != "...")) {
+      s_labels <- paste0(x$sigma$labels[[i]], collapse = " ")
+      header <- c(header, paste0("@labels ", s_labels))
+    }
+    sigma <- c(sigma, "$SIGMA", header, x$sigma$data[[i]], "")  
+  }
+  
+  code <- c(prob, param, init, omega, sigma, x$code, capture)
```
kylebaron, line 304: Capture information pushed to the end of the control stream; handling this explicitly now, like param and init.

kylebaron:
After some fumbling around with tests, I think this is ready to go again.

kylebaron:
A reprex looking at how capture is captured in the yaml file 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(mrgsolve)
#> 
#> 
#>     filter
```

Load a regular mrgsolve model

``` r
mod <- modlib("1005")
#> Loading required namespace: xml2
#> Building 1005 ... done.
mwrite_yaml(mod, "1005.yaml")
x <- readLines("1005.yaml")
cat(x[24:35], sep = "\n")
#>   PERIPH: 0.0
#> capture:
#> - CL
#> - Q
#> - V2
#> - V3
#> - KA
#> - ETA_1 = ETA(1)
#> - ETA_2 = ETA(2)
#> - ETA_3 = ETA(3)
#> - IPRED
#> omega:
```

<sup>Created on 2024-07-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

kylebaron:
This is a bug; I will fix in a PR against this branch for (hopefully) more controlled review. 
``` r
library(mrgsolve)
#> 
#> 
#>     filter
code <- '
$OMEGA 1 2 3
$OMEGA 3 4 5
'
mod <- mcode("foo", code)
#> Building foo ...
#> done.
a <- mwrite_yaml(mod, file = "bar.yaml")
mod <- mread_yaml(file = "bar.yaml")
#> Error in yaml::yaml.load(text): Duplicate map key: '...'
```

<sup>Created on 2024-07-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

tests/testthat/test-mwrite.R
```diff
@@ -0,0 +1,207 @@
...
+
+library(testthat)
+library(mrgsolve)
+
+context("test-mwrite")
+
+test_that("convert list to code", {
+  # This could be simpler, but haven't found an easy way to do `c = "a"`
+  l <- list(a = 1, b = 1.234, c = "a", d = c("x", "y", "z"), e = FALSE)  
+  x <- mrgsolve:::tocode(l)
```
kyleam, line 26: As far as I can tell, this is the only spot in the tests that trigger the call of `.quotes`. If it's not too much trouble, it'd be nice to also test a model that triggers that logic.

tests/testthat/test-mwrite.R
```diff
@@ -0,0 +1,207 @@
...
+  mod4 <- mread_yaml(
+    temp, compile = FALSE, capture = "KA"
+  )
+  expect_true("KA" %in% outvars(mod4)$capture)
+})
+
+test_that("yaml_to_cpp", {
+  mod <- modlib("popex", compile = FALSE)
+  temp <- tempfile()
+  mwrite_yaml(mod, temp)
```
kyleam, line 102: This spot (line 102) will fail if the yaml package isn't installed, so this `testthat` block should call `skip_if_not_installed("yaml")` like other spots in the tests.

Here are additional `mwrite_yaml` calls in the tests that aren't guarded:

-   [tests/testthat/test-mwrite.R:119](https://github.com/metrumresearchgroup/mrgsolve/blob/aa94b84cbaaa5efba9e385b9f6b2897e70509f20/tests/testthat/test-mwrite.R#L119)
-   [tests/testthat/test-mwrite.R:132](https://github.com/metrumresearchgroup/mrgsolve/blob/aa94b84cbaaa5efba9e385b9f6b2897e70509f20/tests/testthat/test-mwrite.R#L132)
-   [tests/testthat/test-mwrite.R:144](https://github.com/metrumresearchgroup/mrgsolve/blob/aa94b84cbaaa5efba9e385b9f6b2897e70509f20/tests/testthat/test-mwrite.R#L144)

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+  l$sigma <- list()
+  l$sigma$data <- lapply(as.list(smat(x)), get_upper_tri)
+  l$sigma$labels <- labels(smat(x))
+  l$sigma$names <- names(smat(x))
+  if(length(l$sigma$data)) {
+    names(l$sigma$data) <- paste0("matrix", seq_along(l$sigma$data))
+    names(l$sigma$labels) <- paste0("matrix",seq_along(l$sigma$labels))
+  }
+  # Other
+  l$env <- as.list(x@envir)
```
kyleam, line 96: Would `envir` be better for consistency?

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+  l$update$hmin <- x@hmin
+  l$update$maxsteps <- x@maxsteps
+  l$update$mxhnil <- x@mxhnil
+  l$update$ixpr <- x@ixpr
+  l$update$mindt <- x@mindt
+  l$update$digits <- x@digits
+  l$update$tscale <- x@tscale
+  l$update$outvars <- unlist(outvars(x), use.names = FALSE)
+  
+  code <- gsub("\\t", "  ", x@code, perl = TRUE)
```
kyleam, line 119: nit: could used `fixed = TRUE` here

```R
code <- gsub("\t", "  ", x@code, fixed = TRUE)
```

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+    code$PROB <- NULL
+  }
+  
+  clob <- c("PARAM", "INPUT", "THETA", "CMT", "INIT", "OMEGA", "SIGMA", 
+            "NMEXT", "NMXML", "VCMT", "SET", "CAPTURE")
+  for(block in clob) {
+    while(block %in% names(code)) {
+      code[[block]] <- NULL
+    }
+  }
```
kyleam, line 151: I think this could be simplified to

```R
code[[clob]] <- NULL
```

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+#' @md
+#' @export
+mwrite_yaml <- function(x, file, digits = 8) {
+  require_yaml()
+  l <- mwrite_model_to_list(x)
+  l$format <- "yaml"
+  out <- yaml::as.yaml(l, precision = digits)
+  if(is.character(file)) {
+    writeLines(con = file, out)  
+  }
```
kyleam, line 223: This `writeLines` call adds an additional newline compared to what `write_yaml` would produce. I think it'd be good to align those.

You could do that by tweaking the `sep` value:

```R
tfile <- tempfile()
x <- list(one = 1, two = 2)

yaml::write_yaml(x, tfile)
readChar(tfile, file.size(tfile))
#> [1] "one: 1.0\ntwo: 2.0\n"

writeLines(yaml::as.yaml(x), tfile)
readChar(tfile, file.size(tfile))
#> [1] "one: 1.0\ntwo: 2.0\n\n"

writeLines(yaml::as.yaml(x), tfile, sep = "")
readChar(tfile, file.size(tfile))
#> [1] "one: 1.0\ntwo: 2.0\n"
```

Or perhaps just switch to using `write_yaml`?

```R
if(is.character(file)) {
  l$format <- "yaml"
  yaml::write_yaml(l, file = file, precision = digits)
}
```

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+}
+
+#' Write a model to native mrgsolve format
+#' 
+#' Model code is written to a file in native mrgsolve format. This 
+#' can be useful for (1) breaking connection to NONMEM modeling outputs that 
+#' are imported by `$NMXML` or `$NMEXT` and (2) saving model updates (e.g., 
+#' an updated parameter list). Models can be read back using [mread()].
+#' 
+#' @inheritParams mwrite
```
kyleam, line 236: This should inherit from `mwrite_yaml` instead of `mwrite`, I think.

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+#' x <- mwrite_cpp(mod, file = temp)
+#' 
+#' mod <- mread(x$file, compile = FALSE)
+#' 
+#' mod
+#' 
+#' @seealso [mwrite_yaml()], [yaml_to_cpp()]
+#' 
+#' @md
+#' @rdname mwrite
```
kyleam, line 260: Drop this `rdname` and just produce `mwrite_cpp.Rd` (consistent with `mwrite_yaml` mapping to `mwrite_yaml.Rd`)?

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+#' mod
+#' 
+#' cppfile <- yaml_to_cpp(temp)
+#' 
+#' readLines(cppfile)
+#' 
+#' @return 
+#' A model object. 
+#' 
+#' @seealso [mwrite_yaml()], [yaml_to_cpp()]
```
kyleam, line 313: Drop `yaml_to_cpp` because it shares this same doc (`mread_yaml.Rd`).

kyleam (approved):
This is looking very nice, though some of the finer details are a bit beyond me. I've left a few comments for your consideration, but nothing major that should hold this up if you disagree.

tests/testthat/test-mwrite.R
kylebaron: Thanks; fixed these.

R/mwrite.R
kylebaron: Good call.

R/mwrite.R
kylebaron: Thanks; didn't know you could do that.

R/mwrite.R
kylebaron: Done.

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
...
+#' 
+#' @md
+#' @export
+mread_yaml <- function(file, model = basename(file), project = tempdir(), ...) {
```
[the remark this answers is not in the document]
kylebaron: Good idea.

R/mwrite.R
kylebaron: @kyleam can you help me with this one? I'm seeing this:

``` r
clob <- c("OMEGA", "SET", "CMT")

code <- list(a = 5, OMEGA = 3, SET = 4, CMT  = 1, SET = "A", CMT = 3, CMT = 5, 
             b = 4, c = 3)
```

# Not working

``` r
code[clob]
#> $OMEGA
#> [1] 3
#> 
#> $SET
#> [1] 4
#> 
#> $CMT
#> [1] 1
try(code[[clob]])
#> Error in code[[clob]] : recursive indexing failed at level 2
```

# What about this?

``` r
w <- which(names(code) %in% clob)
if(length(w)) {
  code <- code[-w]
}
code
#> $a
#> [1] 5
#> 
#> $b
#> [1] 4
#> 
#> $c
#> [1] 3
```

<sup>Created on 2024-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/mwrite.R
```diff
@@ -0,0 +1,415 @@
+.quotes <- function(x) paste0("\"", x, "\"")
```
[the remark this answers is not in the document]
kylebaron: I think I'm going to leave this as is since, as you note, there is really limited inputs that we could take here.

R/mwrite.R
kyleam: Sorry for the typo :/ I meant `[`.  

But what I didn't appreciate was that there could be multiples.  So either your original code or what you suggest work for me.  Sorry for the noise.

kylebaron: Gotcha; thanks. I'll go with the new version since it seems more efficient.

PR #1193, opened 2024-05-21
Refactor dropped non-numeric columns with recover

kylebaron:
# Summary

### DIFF
--- a/R/mrgindata.R
+++ b/R/mrgindata.R
@@ -87,7 +87,7 @@ signal_drop <- function(dm, x, to_signal, context) {
   for(d in drop) {
     type <- paste0(class(x[[d]]), collapse = ",")
     msg <- c(context, " dropped column: ", d, " (", type, ")")
-    message(msg)  
+    warning(msg, call. = FALSE)  
   }
   invisible(NULL)
 }
--- a/R/mrgsolve.R
+++ b/R/mrgsolve.R
@@ -582,6 +582,8 @@ do_mrgsim <- function(x,
       data$.data_row. <- join_data$.data_row.
       carry.recover <- ".data_row."
       drop <- names(which(!is.numeric(join_data)))
+      # Will be dropped with warning when validating data
+      drop <- drop[!drop %in% Pars(x)]
       data <- data[,setdiff(names(data),drop),drop=FALSE]
     }
     recover_idata <- intersect(recover,names(idata))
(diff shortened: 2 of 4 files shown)


R/mrgindata.R
```diff
@@ -81,45 +81,62 @@ convert_character_cmt <- function(data, mod) {
...
-##' @param x data.frame or matrix
-##' @param m a model object
-##' @param verbose logical
-##' @param quiet if \code{TRUE}, messages will be suppressed
+##' @param x data.frame or matrix.
+##' @param m a model object.
+##' @param verbose logical.
+##' @param quiet if `TRUE`, messages will be suppressed.
+##' 
+##' @details
```
kylebaron, line 115: This is mostly called under the hood; but I filled out details since we're referencing the function name in the error message and probably good to get more of this written down somewhere.

R/mrgindata.R
```diff
@@ -81,45 +81,62 @@ convert_character_cmt <- function(data, mod) {
...
+##' An error will be issued when
+##' - non-numeric data is found in columns sharing names with model parameters
+##' - non-numeric data is found in reserved data items related to dosing 
+##'   (see `mrgsolve:::GLOBALS$CARRY_TRAN`)
```
kylebaron, line 119: Is referencing this internal `GLOBALS` data structure a bad idea? I don't want to have to maintain this list off the contents of what is in there but also feels odd to expose this internal data.

R/mrgsolve.R
```diff
@@ -582,6 +582,8 @@ do_mrgsim <- function(x,
       data$.data_row. <- join_data$.data_row.
       carry.recover <- ".data_row."
       drop <- names(which(!is.numeric(join_data)))
+      # Will be dropped with error later when validating data
+      drop <- drop[!drop %in% c(Pars(x), c(Pars(m), GLOBALS$CARRY_TRAN)]
       data <- data[,setdiff(names(data),drop),drop=FALSE]
```
kylebaron, line 587: Honestly can't remember why we're dropping data columns here when it will happen later. For now, will keep it as is.

R/mrgindata.R
kyleam: In my opinion it's okay to reference.  It's useful to those that want to see the list, but it being internal should make it clear that it shouldn't be relied on in scripts, etc.

R/mrgindata.R
```diff
@@ -81,45 +81,62 @@ convert_character_cmt <- function(data, mod) {
...
   }
-  invisible(NULL)
+  abort(
+    message = "Found input data that cannot be used for simulation", 
+    body = body, 
+    call = caller_env()
+  )
 }
 
 ##' Validate and prepare a data sets for simulation
```
kyleam, line 104: Nearby typo in case you want to fix it up here: should be either "prepare data sets" (that's consistent with `valid_idata_set`) or "prepare a data set".

kyleam (approved):
LGTM, left one comment about adjacent doc typo

R/mrgindata.R
kylebaron: Thanks, @kyleam . FIxed now.

PR #1197, opened 2024-05-27
Error when k10 is equal to ka in one-compartment closed-form model

kylebaron:
For closed-form solution to one-compartment model, error when KA and CL/V are too close together. 

- [x] Write test
- [x] Decide on eps

### DIFF
--- a/src/odeproblem.cpp
+++ b/src/odeproblem.cpp
@@ -380,8 +380,8 @@ void odeproblem::advan2(const double& tfrom, const double& tto) {
   
   if(k10 <= 0) Rcpp::stop("k10 has a 0 or negative value");
   // https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
-  // Going for the simpler method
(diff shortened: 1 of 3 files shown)


src/odeproblem.cpp
```diff
@@ -380,6 +380,12 @@ void odeproblem::advan2(const double& tfrom, const double& tto) {
   
   if(k10 <= 0) Rcpp::stop("k10 has a 0 or negative value");
   
+  // Going for the simpler method; using epsilon of ~  100 * DBL_EPSILON
+  // https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
+  if(std::fabs(ka-k10) < 2.22e-14) {
```
kyleam, line 385: I think that approach makes sense here.

---

Just a note:`isSymmetric.matrix` uses the `100 * DBL_EPSILON` as its default value.

https://github.com/wch/r-source/blob/9d01d16b4cda7b53b9a2c94a1e3f078faf1b9eaa/src/library/base/R/eigen.R#L22

kyleam (approved):

PR #1202, opened 2024-06-22
Insert new modeled event records rather than push and sort

kylebaron:
This follows on previous work in #1187 

In this PR, we `insert` event records that get _scheduled_ from within the model rather than push and sort.

### DIFF
--- a/src/devtran.cpp
+++ b/src/devtran.cpp
@@ -663,7 +663,7 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
               mt[mti].now = false;
             }
           }
-          // If the event is stil happening now
+          // If the event is still happening now
           if(mt[mti].now) {
             new_ev->time(tto);
             new_ev->implement(&prob);
@@ -676,8 +676,7 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
                                      new_ev->rate(), 
                                      -299, 
                                      id);
-              a[i].push_back(evoff);
-              std::sort(a[i].begin()+j+1,a[i].end(),CompRec());                       
+              insert_record(a[i], j, evoff, true);
             }
           } else {
             bool do_mt_ev = true;
@@ -687,8 +686,7 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
               do_mt_ev = do_mt_ev && !found;
             }
             if(do_mt_ev) {
-              a[i].push_back(new_ev);
-              std::sort(a[i].begin()+j+1,a[i].end(),CompRec());
+              insert_record(a[i], j, new_ev, put_ev_first);
               mtimehx.push_back(new_ev);
             } 
           }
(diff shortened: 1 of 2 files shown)


src/devtran.cpp
```diff
@@ -676,8 +676,7 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
                                      new_ev->rate(), 
                                      -299, 
                                      id);
-              a[i].push_back(evoff);
-              std::sort(a[i].begin()+j+1,a[i].end(),CompRec());                       
+              insert_record(a[i], j, evoff, true);
```
kyleam, line 679: I'm getting a bit stuck on value for the `put_ev_first` argument here and on line 689.  Before, both sites had

```
    std::sort(a[i].begin()+j+1,a[i].end(),CompRec());
```

`CompRec` sorts by time and then position, so am I correct that it would leading to `put_ev_first=false` behavior (because `push_back()` would have put the new ev in the end position)?

So, what am trying to understand:

 * is there a change in behavior here, and, if so, does it matter?

 * why does this site (line 679) use `true` for the `put_ev_first` argument while the site below (line 689) uses the `put_ev_first` value (which is defined upstream based on `recsort`, lines 123-140)?

kylebaron: It mimics this: 

https://github.com/metrumresearchgroup/mrgsolve/blob/fbc12ece16aa9da589dbf57f812bdc8a888fc0f6/src/devtran.cpp#L576-L590

When an infusion is turned off, we always do that first. But I think that has no consequence on the simulation output. If you did have a record at the same time, turning an infusion off wouldn't change the values in any compartment for records at the same time. It would matter if we were bolus dosing.

kylebaron: This is the relevant bit to look at (the subject of your question): 
https://github.com/metrumresearchgroup/mrgsolve/blob/fbc12ece16aa9da589dbf57f812bdc8a888fc0f6/src/devtran.cpp#L668-L680


We give a dose at the top (`implement()`) and then if it is an infusion, we also have to figure out when the infusion ends and put in a new record for that. The user doesn't really have a say in that but, as I said before, it doesn't matter what happens first. But we're consistent in how it happens in case the user is looking for the infusion end (that would be a fringe case ... but anyway we decide that).

kylebaron: On 689, we are scheduling that in the future (not happening _now_ like what happens on 669 and 679; in this case, I want the order to respect what the user passed for `recsort`

https://github.com/metrumresearchgroup/mrgsolve/blob/fbc12ece16aa9da589dbf57f812bdc8a888fc0f6/src/devtran.cpp#L689

kyleam: Thanks!  That's very helpful.

kyleam (approved):

src/devtran.cpp
kylebaron: Thanks for reviewing this with a critical eye.

PR #1203, opened 2024-06-22
Add replace() function to evtools

kylebaron:
# Summary

To date, the `evtools` plugin has been able to quickly create and give doses given as `bolus()` or `infuse()`. 

This PR implements an equivalent function to `replace()` the amount in a specific compartment with a new amount. This is accomplished via `evid 8` (already-existing functionality). 

The basic setup is similar to `bolus()` and `infuse()`: we provide two overloaded `replace()` functions, one that gets the `self` object first (this will create and "send" the dose in one function call) and another `replace()` function that will create the object and return it to the user for further config.

### DIFF
--- a/inst/base/mrgsolv.h
+++ b/inst/base/mrgsolv.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2013 - 2023  Metrum Research Group
+// Copyright (C) 2013 - 2024  Metrum Research Group
 //
 // This file is part of mrgsolve.
 //
@@ -49,7 +49,7 @@ protected:
 }; 
 
 struct evdata {
-  evdata(double a_, int b_) :  time(a_), evid(b_) {
+  evdata(double a_, int b_) : time(a_), evid(b_) {
     cmt = 1;
     amt = 0.0;
     rate = 0.0;
--- a/inst/base/mrgsolve-evtools.h
+++ b/inst/base/mrgsolve-evtools.h
@@ -34,6 +34,21 @@ void infuse(databox& self, const double amt, const int cmt, const double rate) {
   return;
 }
 
+mrgsolve::evdata replace(const double amt, const int cmt) {
+  mrgsolve::evdata ev(0, 8); 
+  ev.amt = amt; 
+  ev.cmt = cmt;
+  ev.now = true;
+  ev.check_unique = false;
+  return ev;
+}
+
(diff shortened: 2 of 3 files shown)


inst/base/mrgsolv.h
```diff
@@ -49,7 +49,7 @@ struct resim {
 }; 
 
 struct evdata {
-  evdata(double a_, int b_) :  time(a_), evid(b_) {
+  evdata(double a_, int b_) : time(a_), evid(b_) {
```
kylebaron, line 52: Just a lint.

inst/base/mrgsolve-evtools.h
```diff
@@ -34,6 +34,21 @@ void infuse(databox& self, const double amt, const int cmt, const double rate) {
   return;
 }
 
+mrgsolve::evdata replace(const double amt, const int cmt) {
+  mrgsolve::evdata ev(0, 8);
```
kylebaron, line 38: The signature for the constructor is `time` and `evid`.  We use `8` here because that signals to zero out the compartment and add the new amount.

inst/maintenance/unit-cpp/test-evtools.R
```diff
@@ -103,3 +113,28 @@ test_that("evtools - give timed dose now", {
...
+test_that("evtools - replace", {
+  mod <- init(mod, C = 100)
+  mod <- update(mod, delta = 0.1)
+  a <- mrgsim(mod, param = list(mode = 7))
+  before <- filter(a, time < 5)
+  expect_true(all(before$C==100))
+  after <- filter(a, time >= 5)
+  expect_true(all(after$C==50))
+  
+  #' When the replacement is timed into the future, we see the
```
kyleam, line 126: extreme nit-pick: I think it'd be better to use plain comments rather than roxygen-style ones. (I've been allergic to unnecessary roxygen-style comments ever since I spent time debugging [this issue](https://github.com/rstudio/rstudio/issues/12772).)

kyleam (approved):
Nice

inst/maintenance/unit-cpp/test-evtools.R
kylebaron: Thanks, @kyleam; I reverted the comments and will try to stick to plain R comments going forward (I _think_ I do most of the time, but I'm sure there's tons of `#'` comments in the code too.)

PR #1222, opened 2024-08-01
Add reset functions to evtools

kylebaron:
# Summary

Trying to follow a similar pattern to what we do for replace, bolus and friends; this should (hopefully) all be natural / consistent riffing on the `evtools` machinery that is already in place. 

- Reset sends `EVID 3`; it resets the entire PK system (all compartments)
- We have to use some value for compartment; keeping that at the default value of 1 
- Providing `reset()` functions with `self` as the first argument (the event is sent off right away, no return) and without `self` as first argument (you get the object back, with opportunity to interact with it)
- Also providing `reset()` with option to pass along dosing information; this is `EVID 4` in NONMEM and mrgsolve; it resets the system and then doses
- For `reset()` with dose, I'm tacking on `rate` as an optional argument, defaulting to `0` (or bolus); this is a slight difference from what we've done before in that we don't provide separate functions to reset and bolus or reset and infuse
- You can always ask for an infusion dose, update the evid to 4 and get the same result 

I also went back into some of the previously-developed functions and used explicit `evt::` references to functions inside the `evtools` namespace. I got a little spooked adding a function with a vanilla name like `reset()` and wanted to get explicit about all of this in the code. 

## Tests

You can see from Files Changed that tests are in `inst/maintenance/unit-cpp/test-evtools.R`. This can be run with `make test-cpp` from the command line. 

The tests are a little tedious, trying to figure out when the system was reset and when there was a dose. I thought to run the equivalent simulation with no evtools involvement and verify results later on in the process; I think this is pretty compelling, but kept the more brute force tests anyway; probably still useful for catching something that might go wrong some day.

### DIFF
--- a/inst/maintenance/unit-cpp/test-evtools.R
+++ b/inst/maintenance/unit-cpp/test-evtools.R
@@ -61,26 +61,26 @@ if(mode==8 && givedose) {
   evt::retime(rep, reptime);
   self.push(rep);
 }
-if(mode==9 && givedose) {
+if(mode==9 && givedose) { // reset
   evt::reset(self);
-}
-if(mode==10 && givedose) {
+} 
+if(mode==10 && givedose) { // reset, non-self
   evt::ev res = evt::reset();
   res.time = 15;
   self.push(res);
 }
-if(mode==11 && givedose) {
+if(mode==11 && givedose) { // reset with bolus
   evt::reset(self, 100, 3);
 }
-if(mode==12 && givedose) {
+if(mode==12 && givedose) { // reset with infusion
   evt::reset(self, 100, 1, 20);
 }
-if(mode==13 && givedose) {
+if(mode==13 && givedose) { // reset with bolus, non-self
   evt::ev res = evt::reset(100, 3);
   self.push(res);
 }
-if(mode==14 && givedose) {
-  evt::ev res = evt::reset(100, 1, 20);
+if(mode==14 && givedose) { // reset with infusion, non-self
+  evt::ev res = evt::reset(100, 1, 100.0/5.0);
   self.push(res);
 }
 '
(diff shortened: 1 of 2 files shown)


inst/maintenance/unit-cpp/test-evtools.R
```diff
@@ -138,3 +161,97 @@ test_that("evtools - replace", {
...
+  # Dose in to C at time==5
+  x <- filter(out, time >= 5)
+  expect_true(all(x$C==100))
+
+  # Identical results if the self method called or non-self 
+  out1 <- mrgsim(mod, param = list(mode = 11))
+  out2 <- mrgsim(mod, param = list(mode = 13))
+  expect_identical(as.data.frame(out1), as.data.frame(out2))
+    
+  # Do an equivalent simulation with no evtools
```
kyleam, line 222: Nice.

kyleam (approved):
> I got a little spooked adding a function with a vanilla name like `reset()` and wanted to get explicit about all of this in the code.

Makes sense.

----

> We have to use some value for compartment; setting that to 1

Checking my understanding.  The compartment is not explicitly set to 1 in  ([reset](https://github.com/metrumresearchgroup/mrgsolve/blob/54708e72d686423eeed0d603b20acf83ade26d65/inst/base/mrgsolve-evtools.h#L53-L55)), correct?  It's just that it's left at the [default value](https://github.com/metrumresearchgroup/mrgsolve/blob/54708e72d686423eeed0d603b20acf83ade26d65/inst/base/mrgsolv.h#L53).

kylebaron:
> The compartment is not explicitly set to 1 in [reset](https://github.com/metrumresearchgroup/mrgsolve/blob/54708e72d686423eeed0d603b20acf83ade26d65/inst/base/mrgsolve-evtools.h#L53-L55), correct? It's just that it's left at the [default value](https://github.com/metrumresearchgroup/mrgsolve/blob/54708e72d686423eeed0d603b20acf83ade26d65/inst/base/mrgsolv.h#L53).

Yeah, that's right. The description I left was misleading on intent - I didn't have it clear in my own head at first. But now I think that's the only reasonable thing to do; and it's consistent with how `ev()` objects operate on the R side. 

Thanks for reviewing this.

PR #1227, opened 2024-09-01
Fill out evtools function set for dosing

kylebaron:
# Summary 

These functions provide _new_ functionality to (1) gives doses at steady state and (2) give additional doses. We also provide  the user some of the "missing" api for working with these event objects in C++. 

Note: working on this feature, I discovered behavior in #1229 that will need to be addressed in another work stream; it's a known issue now and will need some reimagining of how this in-model dosing is done.

All of the tests compare doses set up via the data set (or event object) and doses that are issued from inside the model. For each combination, there is a test for when the parent event is done "now" and another test when the parent dose is done "later".

**New data members** in `evdata`:
- `ss`
- `ii`
- `addl`

All of these items naturally initialize to `0`. There is no new central functionality needed to handle these items; they just get passed into existing machinery.  

**New functions**: 
- `ss()` - advance the system to steady state prior to diving the dose
- `ii()` - specify the dose interval
- `addl()` - give additional doses
- `amt()` - set / reset the dose amount
- `rate()` - set / reset the infusion rate
- `cmt()`-  set / reset the dose compartment

**Existing functions**: 
- `bolus()` - bolus dose
- `infuse()` - start an infusion
- `reset()` - reset the system or reset and dose
- `replace()` - replace the amount in a compartment (evid 8)
- `retime()` - sets the time attribute _and_ forces `now` to `false`
- `now()` - make an event happen "now"

The goal is to have a fairly complete set of functions to deal with these objects. 

# Example

``` r
library(mrgsolve)
#> 
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 


code <- '
$PLUGIN evtools autodec

$PARAM 
CL = 1, V = 25, KA = 1.3
Dose = 100, Ii = 24, End = 240, Ss = -1

$PKMODEL cmt = "A1,A2", depot = TRUE

$ERROR
if(NEWIND <= 1) {
  evt::ev dose = evt::bolus(Dose, 1);
  evt::ii(dose, Ii); 
  evt::addl(dose, (End/Ii));
  if(Ss > 0) evt::ss(dose, 1);
  self.push(dose);
}
'
mod <- mcode('foo', code)
#> Building foo ...
#> done.
```

# Sensitivity analysis on dose, ii and addl

``` r
idata <- expand.idata(Dose = c(50, 100, 200), Ii = c(6, 12, 24), End = c(120, 240, 360))
```

``` r
out <- mrgsim(mod, idata = idata, end = 360, recover = "Dose,Ii,End", recsort = 3) 

plot(out, A2~time|factor(End)*factor(Ii))
```

# Simulation at steady-state

``` r
idata <- expand.idata(Dose = 100, Ii = c(6, 12, 24), Ss = c(0,1), End = 240)
```

``` r
out <- mrgsim(mod, idata = idata, end = 360, recover = "Dose,Ii,Ss", recsort = 3) 

plot(out, A2~time|factor(Ss))
```

<sup>Created on 2024-09-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: mrgsolve
 Title: Simulate from ODE-Based Models
-Version: 1.5.1
+Version: 1.5.1.9000
 Authors@R: 
     c(person(given = "Kyle T", family = "Baron",
              role = c("aut", "cre"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
+# mrgsolve (development version)
+
 # mrgsolve 1.5.1
 
 - Fixed `yaml_to_cpp()` example code to prevent saving the file to the working 
--- a/inst/base/mrgsolv.h
+++ b/inst/base/mrgsolv.h
@@ -53,7 +53,9 @@ struct evdata {
     cmt = 1;
     amt = 0.0;
     rate = 0.0;
+    ii = 0.0;
     ss = 0; 
+    addl = 0;
     now = false;
     check_unique = false;
   } 
@@ -62,7 +64,9 @@ struct evdata {
   int cmt;
   double amt; 
   double rate;
+  double ii; 
   int ss; 
+  int addl;
   bool now;
   bool check_unique;
 }; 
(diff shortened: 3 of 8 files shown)


kyleam (approved):
Thanks for the detailed summary.

> All of the tests compare doses set up via the data set (or event object) and doses that are issued from inside the model.

These tests look great.

PR #1234, opened 2024-09-20
Refactor autodec cpp variable discovery

kylebaron:
# Summary 

First, when the `autodec` plugin is invoked, mrgsolve discovers and declares variables that the user creates. This PR limits the scope of what is discovered by excluding any variable that has a type of the form `namespace::type` (e.g. `evt::ev`). 

EDIT: 
Discussion with @kyleam 

- Keep going with new code to parse out prefix from autodec discovery
- We'll continue to exclude `namespace::type` declarations
- We'll continue to narrow the autodec code to only work on variables that weren't declared by the user (no change in function)
- Going forward, tighten up autodec search to bypass _anything_ that has _any_ prefix; we'll do that in a stage or two

### DIFF
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -431,10 +431,6 @@ token_space <- function(x) {
 }
 
 get_c_vars2 <- function(y,context) {
-  fors <- grep("\\bfor\\s*\\(", y, perl = TRUE)
-  if(length(fors)) {
-    y <- y[-fors]  
-  }
   m <- gregexpr(move_global_re_find,y,perl=TRUE)
   regm <- unlist(regmatches(y,m))
   if(length(regm)==0) return(data.frame())
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -353,8 +353,6 @@ modelparse_rmd <- function(txt, split=FALSE, drop_blank=TRUE,
   return(ans)
 }
 
-
-
 ## ----------------------------------------------------------------------------
 ## New function set for finding double / bool / int
 ## and moving to global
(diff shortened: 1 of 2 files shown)


R/modspec.R
```diff
@@ -433,6 +431,10 @@ token_space <- function(x) {
 }
 
 get_c_vars2 <- function(y,context) {
+  fors <- grep("\\bfor\\s*\\(", y, perl = TRUE)
```
kylebaron, line 434: Exclude lines that are for loops.

R/modspec.R
```diff
@@ -865,9 +867,28 @@ autodec_find <- function(code) {
   if(length(code)==0) {
     return(NULL)
   }
-  ans <- regmatches(code, regexpr("[._[:alnum:]]+ *=([^=]|$)", code, perl = TRUE))
-  ans <- unique(sub(" *=.?$", "", ans, perl = TRUE))
-  ans[!grepl(".", ans, fixed = TRUE)]
+  m <- regexpr("[._[:alnum:]]+ *=([^=]|$)", code, perl = TRUE)
```
kylebaron, line 870: Narrow the search for autodec.

R/modspec.R
```diff
@@ -433,6 +431,10 @@ token_space <- function(x) {
 }
 
 get_c_vars2 <- function(y,context) {
+  fors <- grep("\\bfor\\s*\\(", y, perl = TRUE)
+  if(length(fors)) {
+    y <- y[-fors]  
+  }
```
kylebaron, line 437: Exclude lines that are for loops.

R/modspec.R
kylebaron: This saves the match locations so we can look at what happens from the start of the line to the variable name.

R/modspec.R
```diff
@@ -865,9 +867,29 @@ autodec_find <- function(code) {
...
+  pre <- trimws(pre, which = "left")
+  if(all(pre=="")) {
+    return(unique(ans[!has_dot]))  
+  }
+  pre <- strsplit(pre, "[ )(}{\\[\\]]", perl = TRUE)
+  p0 <- sapply(pre, "[", 1L)
+  pre <- lapply(pre, rev)
+  p1 <- sapply(pre, "[", 1L)
+  p2 <- sapply(pre, "[", 2L)
+  drop1 <- p1 %in% c("double", "int", "bool", "const", "static", "unsigned")
```
kylebaron, line 887: These weren't dropped before, but now they are. Those declarations will get processed as usual when `autodec` _isn't_ invoked.

R/modspec.R
```diff
@@ -865,9 +867,29 @@ autodec_find <- function(code) {
...
+  drop2 <- p2 %in% c("static", "const", "unsigned")
+  drop3 <- grepl("::", p0, fixed = TRUE)
```
kylebaron, line 889: This is what got me into the PR: I noticed that some variables were getting declared globally as `double` and then locally as `evt::ev`. This isn't affecting any functionality, but it's not correct and those variables show up as getting declared as double if you look.

tests/testthat/test-modspec.R
```diff
@@ -434,8 +438,16 @@ test_that("autodec parsing", {
   expect_equal(x, character(0))
   x <- mrgsolve:::autodec_find("if(TIME != 1 ) {")  
   expect_equal(x, character(0))
+  x <- mrgsolve:::autodec_find("if(TIME != 1 ) {ccc = 11;")  
+  expect_equal(x, "ccc")
```
kylebaron, line 442: Making sure we don't detect `{` in this one.

kyleam (approved):
Stepping through the `autodec_find` for various cases, I didn't spot any issues and the test cases look good.

> Going forward, tighten up autodec search to bypass _anything_ that has _any_ prefix; we'll do that in a stage or two

I think that will be a very nice change from a complexity standpoint.

Issue #1182, opened 2024-03-22
Fail to detect missing differential equation with autodec and nm-vars plugin


kylebaron:
Looks like you have to be using both

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

code <- '
$PLUGIN autodec nm-vars
$CMT D B C2
$DES
dxdt_D = -1; 
dxdt_B = 5;
dxdt_C = 5;
'

mod <- mcode("foo", code)
#> Building foo ...
#> done.


mrgsim(mod)
#> Model:  foo 
#> Dim:    25 x 5 
#> Time:   0 to 24 
#> ID:     1 
#>     ID time  D  B C2
#> 1:   1    0  0  0  0
#> 2:   1    1 -1  5  0
#> 3:   1    2 -2 10  0
#> 4:   1    3 -3 15  0
#> 5:   1    4 -4 20  0
#> 6:   1    5 -5 25  0
#> 7:   1    6 -6 30  0
#> 8:   1    7 -7 35  0
```

<sup>Created on 2024-03-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kylebaron:
If you drop the plugin:
```
``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

code <- '
$CMT D B C2
$DES
dxdt_D = -1; 
dxdt_B = 5;
dxdt_C = 5;
'

mod <- mcode("foo", code)
#> Warning: Missing differential equation(s):
#> --| missing: C2
#> --| suppress with @!audit block option
#> Building foo ...
#> error.
#> 
#> ---:: stderr ::---------------------------------------------
#> using C++ compiler: ‘Apple clang version 14.0.3 (clang-1403.0.22.14.1)’
[...]

PR #1235, opened 2024-09-22
Refactor audit

kylebaron:
# Summary

When a model contains ODE, we need an equation written for every compartment in `$ODE` block. "audit" means we look for some evidence that every compartment has been referenced somewhere.

The audit code was totally different when `nm-vars` plugin was invoked compared to when it wasn't invoked. 
- `nm-vars`: ODE use `DADT(n) = ` on the lhs
- non-`nm-vars`: ODE use `dxdt_CMT = ` on the lhs

The code for checking in the two different approaches was different because there is special handling for `nm-vars` style and it was a little easier at the time to just keep that down in the code where all the `nm-vars` functionality was handled. 

One bad decision from `nm-vars` is that we would only audit if there was at least one `DADT(n)` found. That wasn't safe because #1182 is a report from a user who invoked `nm-vars`, but used the non-`nm-vars` style, which is allowed. And of course the name of a `dxdt_CMT` variable was misspelled. The missing ODE wasn't detected and we spent some time trying to figure that out. 

So here we centralize checking `$ODE` for the required lhs.


Makefile
```diff
@@ -103,13 +103,15 @@ test2:
 test-cpp: 
 	Rscript -e 'testthat::test_dir("inst/maintenance/unit-cpp")'
 
+remove:
```
kylebaron, line 106: I added this.

R/modspec.R
```diff
@@ -195,24 +195,48 @@ check_spec_contents <- function(x, crump = TRUE, warn = TRUE, ...) {
...
-  cmt <- names(init(x))
-  if(!has_name("ODE", spec) | !warn | length(cmt) ==0) {
+grepl_dxdt_ode <- function(spec, cmt) {
+  dx <- paste0("dxdt_", cmt)
+  z <- vapply(dx, FUN.VALUE = TRUE, function(dxi) {
+    any(grepl(dxi, spec[["ODE"]], fixed = TRUE))
+  })  
+  names(z) <- dx
+  z
+}
```
kylebaron, line 205: Returns a logical vector indicating which ODEs were found.

R/modspec.R
```diff
@@ -195,24 +195,48 @@ check_spec_contents <- function(x, crump = TRUE, warn = TRUE, ...) {
...
+
+audit_spec <- function(x, spec, nmv, env, warn = TRUE) {
```
kylebaron, line 207: `nmv` is an object containing all the information we have about the `nm-vars` in use (e.g. `F2`, A_0(1)`, etc. It is null if `nm-vars` isn't invoked.

R/mread.R
```diff
@@ -357,7 +357,9 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
     spec[["ODE"]] <- c(spec[["ODE"]], paste0("dxdt_", vcmt, "=0;"))
   }
   
-  # Handle nm-vars plugin
+  # Handle nm-vars plugin; initializing `nmv` to `NULL` here for use in 
+  # the audit check later on
+  nmv <- NULL
```
kylebaron, line 362: Initialize `nmv` here so we can pass it to the audit function later.

R/mread.R
```diff
@@ -366,8 +368,7 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
     build[["nm-vars"]] <- nmv
     audit_nm_vars(
       spec,
-      param = param, 
-      init = init, 
+      x,
```
kylebaron, line 371: I simplified this too ... no need to pass both when we can just pass the object `param` and `init` come from.

R/nm-mode.R
```diff
@@ -86,11 +86,12 @@ any_nm_vars <- function(x) {
   list(found_any = length(ans) > 0, match = ans)
 }
 
-audit_nm_vars <- function(x, param, init, build, nmv, env) {
-  bad_param <- any_nm_vars(names(param))
-  bad_init <- any_nm_vars(names(init))
+audit_nm_vars <- function(spec, x, build, nmv, env) {
```
kylebaron, line 89: Signature includes only `x`  rather than `param` and `init`.

R/nm-mode.R
```diff
@@ -86,11 +86,12 @@ any_nm_vars <- function(x) {
...
+  cmt <- Cmt(x)
+  bad_param <- any_nm_vars(Pars(x))
+  bad_init <- any_nm_vars(cmt)
   bad_cpp <- any_nm_vars(build[["cpp_variables"]][["var"]])
```
kylebaron, line 93: Here we get the names from `x` rather than `param` and `init`.

R/nm-mode.R
```diff
@@ -86,11 +86,12 @@ any_nm_vars <- function(x) {
...
-  audit_dadt <- isTRUE(env[["audit_dadt"]]) && length(nmv[["dcmtn"]]) > 0
+  audit_dadt <- isTRUE(env[["audit_dadt"]])
```
kylebaron, line 94: Part of the audit unification; we audit with or with `nm-vars`, as long as user didn't say not to audit via block option.

R/nm-mode.R
```diff
@@ -107,9 +108,8 @@ audit_nm_vars <- function(x, param, init, build, nmv, env) {
     msg <- paste0("--| reserved: ", bad_cpp[["match"]])
     err <- c(err, msg)
   }
-  cmtn <- seq_along(init)
-  if(length(cmtn) > 0) {
-    err <- c(err, audit_nm_vars_range(nmv, cmtn, audit_dadt = audit_dadt))
+  if(length(cmt) > 0) {
+    err <- c(err, audit_nm_vars_range(nmv, cmt))
```
kylebaron, line 112: No longer auditing here when processing `nm-vars`.

R/modspec.R
```diff
@@ -195,24 +195,48 @@ check_spec_contents <- function(x, crump = TRUE, warn = TRUE, ...) {
   return(invisible(NULL))
 }
 
-audit_spec <- function(x, spec, warn = TRUE) {
-  cmt <- names(init(x))
-  if(!has_name("ODE", spec) | !warn | length(cmt) ==0) {
+grepl_dxdt_ode <- function(spec, cmt) {
+  dx <- paste0("dxdt_", cmt)
+  z <- vapply(dx, FUN.VALUE = TRUE, function(dxi) {
+    any(grepl(dxi, spec[["ODE"]], fixed = TRUE))
```
kyleam, line 201: Are partial matches here a concern?  For example

```diff
diff --git a/inst/maintenance/unit/test-audit.R b/inst/maintenance/unit/test-audit.R
index 482f0ec1..1897b667 100644
--- a/inst/maintenance/unit/test-audit.R
+++ b/inst/maintenance/unit/test-audit.R
@@ -38,4 +38,10 @@ test_that("test audit ode block dadt", {
   code6 <- "$CMT @number 3"
   expect_silent(mod <- mcode("audit6", code6, compile = FALSE))
   expect_is(mod, "mrgmod")
+
+  code7 <- "$CMT A \n $ODE \n dxdt_A1  = 0;"
+  expect_warning(
+    mcode("audit7", code7, compile = FALSE, quiet = FALSE), 
+    regexp = "--| missing: dxdt_A"
+  )
 })
```

Perhaps something like this to make it more specific:

```diff
diff --git a/R/modspec.R b/R/modspec.R
index c0be7196..12b1a0c4 100644
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -198,7 +198,8 @@ check_spec_contents <- function(x, crump = TRUE, warn = TRUE, ...) {
 grepl_dxdt_ode <- function(spec, cmt) {
   dx <- paste0("dxdt_", cmt)
   z <- vapply(dx, FUN.VALUE = TRUE, function(dxi) {
-    any(grepl(dxi, spec[["ODE"]], fixed = TRUE))
+    pat <- paste0(" *\\Q", dxi, "\\E", " *=")
+    any(grepl(pat, spec[["ODE"]]))
   })  
   names(z) <- dx
   z
```

Makefile
kyleam: Can't deny that :)

inst/maintenance/unit/test-audit.R
```diff
@@ -20,14 +20,22 @@ test_that("test audit ode block dadt", {
   expect_silent(mcode("audit2b", code2b, compile = FALSE))
   
   code3 <- "$PLUGIN nm-vars \n $CMT A1 B C  \n $ODE @audit \n DADT(1)  = 0;"
-  expect_error(
+  expect_warning(
```
kyleam, line 23: Hmm, so there are some cases in user code where what would have been an error will now just be a warning after this change?

inst/maintenance/unit/test-nm-vars.R
```diff
@@ -239,3 +239,81 @@ test_that("nm-vars no frda items", {
...
+  $ode @!audit
+  dxdt_a = 1; 
+  dxdt_b = 2; 
+  '
+  expect_silent(mcode("nmv-audit-2", code, compile = FALSE))
+  
+  code <- '
+  $plugin nm-vars
+  $cmt a b c
+  $ode @!audit
```
kyleam, line 282: Is the point of these tests to check that you don't get a warning when all the ODEs are specified?  And if so, shouldn't the `@!audit` here and on line 292 be dropped?

Makefile
kylebaron: LOL

R/modspec.R
kylebaron: Good idea; thanks. Tests added for this as well.

kylebaron: I never came across `\Q ... \E` before; pretty cool; will definitely use this.

inst/maintenance/unit/test-audit.R
kylebaron: Yes. I must have been on edge when writing the NM stuff.

What about reversing this and making it an error all the time now?  I was wondering about this ... it probably should be an error and users can opt out from this audit in fringe cases where each compartment isn't explicitly written into the `$ODE` block (like writing in a loop). 

What do you think?

inst/maintenance/unit/test-nm-vars.R
kylebaron: Dropped; thanks.

kylebaron:
@kyleam I didn't see this warning before? https://github.com/metrumresearchgroup/mrgsolve/actions/runs/11194343722/job/31120704962?pr=1235#step:6:57

But can't say I looked either.

tests/testthat/test-modspec.R
```diff
@@ -113,7 +113,7 @@ test_that("multiple blocks allowed or not allowed", {
   )
   expect_is(mod, "mrgmod")
   
-  code <- "$ODE a = 3\n$ODE b = 55\n$PARAM x = 3"
+  code <- "$CMT a\n$ODE dxdt_a = 3\n$ODE b = 55\n$PARAM x = 3"
```
kylebaron, line 116: I made this change to the test, but looking at it again, wondering why it's popping up.

kylebaron:
```r
> code <- "$ODE a = 3\n$ODE b = 55\n$PARAM x = 3"
> mcode("foo", code)
Warning: Could not find a $INIT or $CMT block.
error.

```

kylebaron: Ok; this is fallout from the `warn` / `quiet` business. We make the tests run with `quiet`, so this warning was suppressed before, when `quiet` could also suppress `warn`.  I think the current behavior is correct ... notify user no matter what if there is `$ODE` code but no compartments.  Maybe should be an error; but it's been a warning that the user would normally see from the start.

R/modspec.R
```diff
@@ -196,24 +196,49 @@ check_spec_contents <-  function(x, crump = TRUE, warn = TRUE, ...) {
   return(invisible(NULL))
 }
 
-audit_spec <- function(x, spec, warn = TRUE) {
-  cmt <- names(init(x))
-  if(!has_name("ODE", spec) | !warn | length(cmt) ==0) {
+grepl_dxdt_ode <- function(spec, cmt) {
+  dx <- paste0("dxdt_", cmt)
+  z <- vapply(dx, FUN.VALUE = TRUE, function(dxi) {
+    pat <- paste0(" *\\Q", dxi, "\\E", " *=")
```
kyleam, line 202: Looking at my [suggestion](https://github.com/metrumresearchgroup/mrgsolve/pull/1235#discussion_r1779161968) again, that leading " *" is superfluous.  I'm guessing I meant to add a "^" anchor there, but that would have been a bad idea because, at the very least, things like "$ODE dxdt_a = 3" should match.

As is, the pattern leaves open the possibility of a partial match at the front end (e.g., "bahdxdt_CMT =").  I'm not sure how much of a worry that is, but my updated suggestion is

```diff
diff --git a/R/modspec.R b/R/modspec.R
index 27be48de..88ad80ef 100644
--- a/R/modspec.R
+++ b/R/modspec.R
@@ -199,7 +199,7 @@ check_spec_contents <-  function(x, crump = TRUE, warn = TRUE, ...) {
 grepl_dxdt_ode <- function(spec, cmt) {
   dx <- paste0("dxdt_", cmt)
   z <- vapply(dx, FUN.VALUE = TRUE, function(dxi) {
-    pat <- paste0(" *\\Q", dxi, "\\E", " *=")
+    pat <- paste0("\\b\\Q", dxi, "\\E", " *=")
     any(grepl(pat, spec[["ODE"]]))
   })  
   names(z) <- dx
```

inst/maintenance/unit/test-audit.R
kyleam: Thanks for confirming.  I don't have a strong leaning there, and don't have any objections to going with the warning if you prefer to go with the less aggressive behavior when consolidating the nm-vars and non-nmvars checks.

kyleam (approved):
Thanks for the updates.  Looking good to me.

Left an updated suggestion for the `grepl_dxdt_ode` regex for you consideration.

kyleam:
> @kyleam I didn't see this warning before? https://github.com/metrumresearchgroup/mrgsolve/actions/runs/11194343722/job/31120704962?pr=1235#step:6:57

```
Warning: Dependence on R version ‘3.6.2’ not with patchlevel 0
```

Yeah, I have noticed that showing up under the annotations of the CI summary.  An example from 4 months ago:
https://github.com/metrumresearchgroup/mrgsolve/actions/runs/9762229644

Here are  the relevant docs and comments for this check:

```
@item _R_CHECK_R_DEPENDS_
Check that any dependence on R is not on a recent patch-level version
such as @code{R (>= 3.3.3)} since blocking installation of a package
will also block its reverse dependencies.  Possible values
@samp{"note"}, @samp{"warn"} and logical values (where currently true
values are equivalent to @samp{"note"}).
Default: false (but @samp{"warn"} for @option{--as-cran}).
```

```
## Dependence on say R >= 3.4.3 when 3.4 is current can
## cause problems with revdeps (and did for 3.2.x).
## We only check recent ones: maybe previous two
## (R-release and R-old-release) while this is R-devel
```
https://github.com/wch/r-source/blob/e48723ec1a1a49f62de4b637b8f82439ec6e52d2/src/library/tools/R/check.R#L1205-L1235

Given these are jobs running older R versions, I think we can safely ignore the warning.  We could set `_R_CHECK_R_DEPENDS_=FALSE`  (perhaps just for jobs other than `r: release`) to avoid the warning.

R/modspec.R
kylebaron: @kyleam Thanks; updated.

kylebaron:
warning versus error: if I was doing this new today, I'd probably make it an error. But I think the majority of users have been getting a warning so will keep it at that for now. But I'm going to open an issue about transitioning that at some point ... it's probably worth considering again.

Issue #1204, opened 2024-06-25
Handle multiple PLUGIN blocks


kylebaron:
Warn, error, or collect.  Responding to user mishap when using multiple $PLUGIN blocks.

kylebaron:
This was addressed in a global assessment of which blocks can be dups and making sure documentation is correct.

PR #1230, opened 2024-09-08
Add $EVENT block

kylebaron:
This PR adds a new block (`$EVENT`) which functions exactly like (`$TABLE`), but it gets called before `$TABLE` gets called. The only reason for this is so that dynamic dosing events can happen prior to output calculations in `$TABLE`. 

# Example with `$EVENT`

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

code <- '
$PLUGIN evtools

$CMT B

$MAIN 
double a = 1; 

$ODE
dxdt_B = -0.1*B;

$EVENT
if(TIME==0) {
  evt::ev dose = evt::infuse(100, 1, 10);
  evt::ii(dose, 24); 
  evt::addl(dose, 4);
  evt::ss(dose, 1);
  self.push(dose);
}
a = 2;

$ERROR
capture cp = B/12; 

$CAPTURE a
'

mod <- mcode("test", code, soloc = "build")
#> Creating build directory: build
#> Building test ... done.

out <- mrgsim(mod, end = 196)
plot(out)
```

``` r
out
#> Model:  test 
#> Dim:    197 x 5 
#> Time:   0 to 196 
#> ID:     1 
#>     ID time     B a    cp
#> 1:   1    0 17.14 2 1.429
#> 2:   1    1 25.03 2 2.086
#> 3:   1    2 32.16 2 2.680
#> 4:   1    3 38.62 2 3.218
#> 5:   1    4 44.46 2 3.705
#> 6:   1    5 49.74 2 4.145
#> 7:   1    6 54.53 2 4.544
#> 8:   1    7 58.85 2 4.905

as.list(mod)$cpp_variables
#>      type var context
#> 1  double   a    main
#> 2 capture  cp   table
p <- deparse(mod@shlib$pointers)
```

# Example with `$ERROR`

``` r
code2 <- '
$PLUGIN evtools

$CMT B

$ODE
dxdt_B = -0.1*B;

$ERROR
if(TIME==0) {
  evt::ev dose = evt::infuse(100, 1, 10);
  evt::ii(dose, 24); 
  evt::addl(dose, 4);
  evt::ss(dose, 1);
  self.push(dose);
}

[...]

PR #1238, opened 2024-09-23
Check spec for duplicate blocks and allow multiple $PLUGIN

kylebaron:
This PR adds clarity around where we do and do not allow multiple blocks, adding tests for places where multiples aren't allowed. 

We also now allow multiple $PLUGIN blocks. I don't anticipate this getting used a lot, but allowing multiples here, I think, adds clarity to the list of blocks where we don't allow multiples. 

Note: we'll have to add $EVENT to the list where we don't allow multiples when that PR comes through. 

#1204


kyleam (approved):
Seems like a good idea to me.

R/Aaaa.R
```diff
@@ -82,6 +82,10 @@ block_list <- c("ENV", "PROB", "PARAM", "INIT",
                 "PREAMBLE", "PRED", "BLOCK", "TRANSIT", "YAML", "NMEXT", 
                 "INPUT", "EVENT")
 
+block_list_single <- c("MAIN", "SET", "GLOBAL", "PREAMBLE", "PRED", "PKMODEL", 
+                       "ENV", "CMTN", "INCLUDE", "NAMESPACE", "BLOCK", 
+                       "TRANSIT", "YAML", "EVENT")
```
kylebaron, line 87: Added EVENT here after merging #1230

PR #1246, opened 2024-10-08
General updates to Rd documentation

kylebaron:
Reformatting all documentation.

- argument descriptions start with lower case and end with period.
- switch to `@md` where we make changes to the other parts of the help topic
- change `##'` to `#'`
- Add examples and return values where we can
- Definitely some files I left alone

Latest pkgdown site:[DOCS.zip](https://github.com/user-attachments/files/17337395/DOCS.zip) Might aid review.

### DIFF
--- a/R/chain.R
+++ b/R/chain.R
@@ -86,7 +86,7 @@ req.mrgmod <- function(x,...) {
 #' 
 #' e <- ev(amt = 100, ii = 6, addl = 3, WT = 70, dose = amt)
 #' 
-#' out <- mod %>% carry_out(amt, dose, WT)
+#' out <- mod %>% ev(e) %>% carry_out(amt, dose, WT) %>% mrgsim()
 #' 
 #' head(out)
 #' 
--- a/R/chain.R
+++ b/R/chain.R
@@ -75,11 +75,11 @@ req.mrgmod <- function(x,...) {
 #' @param ... unquoted names of data items to copy into the simulated output.
 #' 
 #' @details
-#' There is also a `carry_out` argument to [mrgsim()] that can 
-#' be set to accomplish the same thing as a call to `carry_out` in 
-#' the pipeline.
+#' There is also a `carry_out` argument to [mrgsim()] that can be set to 
+#' accomplish the same thing as a call to `carry_out` in the pipeline.
 #' 
-#' `carry.out` and carry_out`.  Using the underscore version is now preferred.
+#' `carry.out` and `carry_out` both do the same thing; using the underscore 
+#' version is now preferred.
 #' 
 #' @md
 #' @export
--- a/R/chain.R
+++ b/R/chain.R
@@ -32,7 +32,7 @@
 #' 
 #' Note the difference between `req` and `Req`: the former only 
 #' selects compartments to appear in output while the latter selects both 
-#' compartments and captured items.  Also, when there are items are explicitly
+#' compartments and captured items.  Also, when there are items explicitly
 #' listed in `Req`, all other compartments or captured items not listed
 #' there are ignored.  But when compartments are selected with `req`
 #' all of the captured items are returned.  Remember that `req` is 
@@ -137,7 +137,7 @@ tscale <- function(x,value=1,...) {
 #' @param ... not used.
 #' 
 #' @details
-#' There is also a `obsonly` argument to [mrgsim()] that can 
+#' There is also an `obsonly` argument to [mrgsim()] that can 
 #' be set to accomplish the same thing as a call to `obsonly()` in 
 #' the pipeline.
 #' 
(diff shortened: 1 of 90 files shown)


_pkgdown.yml
```diff
@@ -1,6 +1,7 @@
 template:
   bootstrap: 5
 title: mrgsolve
+url: https://github.com/metrumresearchgroup/mrgsolve
```
kyleam, line 4: Hmm, isn't this supposed to be the URL of the pkgdown site? At least for [auto-linking](https://pkgdown.r-lib.org/articles/linking.html#across-packages) to work across packages, I think that the value in `_pkgdown.yml` should be `https://mrgsolve.org/docs` and then in `DESCRIPTION` the `URL` field should be

```
URL: https://mrgsolve.org/docs, https://github.com/metrumresearchgroup/mrgsolve
```

R/chain.R
```diff
@@ -16,34 +16,35 @@
...
+#' @param ... unquoted names of compartments or tabled items.
+#' 
+#' @details
+#' There is also a `Req` argument to [mrgsim()] that can 
+#' be set to accomplish the same thing as a call to `Req` in 
+#' the pipeline.
+#' 
+#' Note the difference between `req` and `Req`: the former only 
+#' selects compartments to appear in output while the latter selects both 
+#' compartments and captured items.  Also, when there are items are explicitly
```
kyleam, line 35: Not new to this PR, but should "when there are items are explicitly listed &#x2026;" be "when there are items explicitly listed &#x2026;" (i.e. drop second "are")?

R/chain.R
```diff
@@ -52,85 +53,95 @@ Req.mrgmod <- function(x,...) {
...
-##' 
-##' @export
+#' Collect only observation records in the simulated output
+#'
+#' @param x model object.
+#' @param value use `TRUE` to collect and return observation records only.
+#' @param ... not used.
+#' 
+#' @details
+#' There is also a `obsonly` argument to [mrgsim()] that can
```
kyleam, line 140: ﻿typo (not new): "a" -> "an" ?

R/class_tgrid.R
```diff
@@ -107,41 +108,42 @@ as_deslist <- function(data, descol="ID") {
...
-##'
-##' plot(out,CP~., type='b')
-##' }
+#' Create a simtime object
+#'
+#' simtime objects allow the user to specify simulation start and end times, 
+#' along with the simulation time step.
+#'
+#' @rdname tgrid
+#' @name tgrid
```
kyleam, line 117: [ not changed with this PR ]

This is one of several spots that use both `@rdname` and `@name` with the same value. In the case, I think `rdname` is unnecessary.

R/data_set.R
```diff
@@ -428,55 +438,57 @@ ev_assign <- function(l, idata, evgroup, join = FALSE) {
...
+#' or every other day (however you want to define that) etc.
+#' 
+#' @param ev an event object.
+#' @param days comma- or space-separated character string of valid days of the
+#' the week (see details).
+#' @param addl additional doses to administer.
+#' @param ii inter-dose interval; intended use is to keep this at the 
+#' default value.
+#' @param unit time unit; the function can only currently handle hours or days.
+#' @param ... event objects named by one the valid days of the week
```
kyleam, line 462: typo (not new to this PR): "by one the" -> "by one of the"

R/data_set.R
```diff
@@ -428,55 +438,57 @@ ev_assign <- function(l, idata, evgroup, join = FALSE) {
...
+#' - `w` for Wednesday
+#' - `th` for Thursday
+#' - `f` for Friday
+#' - `sa` for Saturday
+#' - `s` for Sunday
+#' 
+#' 
+#' The whole purpose of this function is to schedule doses on specific
+#' days of the week, in a repeating weekly schedule.  Please do use caution 
+#' when changing `ii` from it's default value.
```
kyleam, line 478: typo (not new): "it's" -> "its"

kyleam (approved):
The cleanups and extensions look good (and like they took a lot of tedious and careful work).

My inline comments are nearly all about existing typos that I spotted while going through the diff.

> argument descriptions start with lower case and end with period.

Great, I think it's very nice to have consistency within a project.

When there's not a clear project preference (our projects in general are all over the place), I tend to start with an upper case letter to align with the [tidyverse convention](https://style.tidyverse.org/documentation.html#documenting-parameters):

> For most tags, like `@param`, `@seealso` and `@return`, the text should be a sentence, starting with a capital letter and ending with a full stop.

R/chain.R
kylebaron: Right.

R/class_tgrid.R
kylebaron: I think they can both go?  this documentation is associated with a function called `tgrid()`.

_pkgdown.yml
kylebaron: I've been hesitant to list those urls in the package b/c I've been maintaining it and I know CRAN will complain if the url isn't reachable. Well, it's been stable for a while so will just go for it.

PR #1250, opened 2024-10-15
Adding workflow for rhub

kylebaron:
See https://r-hub.github.io/rhub/#setup


kyleam (approved):
Okay, so as far as I understand

 * this is a manual workflow that can be triggered by calling `rhub_check`
 * once you call that, a corresponding section will show up at <https://github.com/metrumresearchgroup/mrgsolve/actions/workflows/rhub.yaml> ([example from cli repo](https://github.com/r-lib/cli/actions/workflows/rhub.yaml))
 * but you'll need to wait for rhub.yaml to hit main before you can actually try that out

kylebaron:
Thanks; I had been using the old rhub interface when testing for CRAN, but looks like you can't do that any more. Will give this a try.

Issue #1261, opened 2025-02-13
System won't go to steady state with EVID=4


kylebaron:
``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

code <- '
$PARAM @annotated
CL   :  1 : Clearance (volume/time)
V2   : 20 : Central volume (volume)
Q    :  2 : Inter-compartmental clearance (volume/time)
V3   : 10 : Peripheral volume of distribution (volume)
KA   :  1 : Absorption rate constant (1/time)

$PARAM lag = 1

$CMT @annotated
EV     : Extravascular compartment (mass)
CENT   : Central compartment (mass)
PERIPH : Peripheral compartment (mass) 

$GLOBAL
#define CP (CENT/V2)

$PKMODEL ncmt = 2, depot = TRUE

$MAIN 
ALAG_EV = lag;

$CAPTURE @annotated
CP : Plasma concentration (mass/time)
'
mod <- mcode("code", code)
#> Building code ...
#> done.
```

# Works with evid==1

``` r
dose <- ev(amt = 100, ii = 6, evid = 1, ss = 1, addl = 3)
mrgsim(mod, dose, start = 0, carry_out = "evid", recsort = 3, start = 0) %>% 
  filter(time %in% c(0,6,12))
#> # A tibble: 4 × 7
#>      ID  time  evid    EV  CENT PERIPH    CP
#>   <dbl> <dbl> <dbl> <dbl> <dbl>  <dbl> <dbl>
#> 1     1     0     1 0.675  315.   167.  15.8
#> 2     1     0     0 0.675  315.   167.  15.8
#> 3     1     6     0 0.675  315.   167.  15.8
#> 4     1    12     0 0.675  315.   167.  15.8
```

# Works with evid==4 and no lag time

``` r
dose <- ev(amt = 100, ii = 6, evid = 1, ss = 1, addl = 3, lag = 0)
mrgsim(mod, dose, start = 0, carry_out = "evid", recsort = 3, start = 0) %>% 
  filter(time %in% c(0,6,12))
#> # A tibble: 4 × 7
[...]

PR #1262, opened 2025-02-22
Correctly handle EVID4 doses at steady state with lag time

kylebaron:
`EVID===4` resets the system and then doses. `SS==1` advances the system to steady-state and then doses.  The current behavior is that `EVID 4` records at steady state act just like `EVID 1` records at steady state; this was confirmed by testing against NONMEMs behavior. 

It was noted in #1261 that, when there is also a lag time for the dose, mrgsolve was dosing as immediately after reset rather than immediately after advancing to steady state.  This PR fixes that behavior. 

Some terminology / background

- A dose that is not `Armed` means there appears to be a dose at that time, but nothing is actually dosed; this is in place at the time of the apparent dose
- When there is a lagged dose, we disarm the apparent dosing record, then schedule a dose or series of doses at the lag time; these are "phantom" records that actually implement a dose some time after the apparent dose

1. In the dose `implement` method, if the record is not `Armed`, we just break out; there is nothing to reset and we don't want to go on and dose after that reset
2. In the main simulation code, lagged doses all lose the ss flag and they are all converted in to evid 1, to act like vanilla dosing records

Confirming that we now match up with NONMEM here:
https://github.com/mrgsolve/nmtests/blob/master/dosing/dosing-vignette.md#23-reset-and-dose-evid-4-with-ss1-and-lag

And the mis-match prior to fixing here:
https://github.com/mrgsolve/nmtests/commit/fc464d8b8af023faff23a9359e773ee29ae3ec51#diff-f9292bfdbc09d870158ed0c0760957e19c4161d00f623e1c4b143ba83503a400

Also, there is a unit test showing that steady state doses with a lag time are identical whether EVID is 1 or 4. 

We should consider if that `implement` method should exit immediately for _any_ unarmed event; maybe after we set `itask`.  See #1267

### DIFF
--- a/mrgsolve.Rproj
+++ b/mrgsolve.Rproj
@@ -1,4 +1,5 @@
 Version: 1.0
+ProjectId: 0063b92e-a050-4fd7-986e-be62f57eb1b1
 
 RestoreWorkspace: Default
 SaveWorkspace: Default
--- a/inst/maintenance/unit/test-alag.R
+++ b/inst/maintenance/unit/test-alag.R
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 - 2019  Metrum Research Group
+# Copyright (C) 2013 - 2025  Metrum Research Group
 #
 # This file is part of mrgsolve.
 #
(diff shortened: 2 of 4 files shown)


kylebaron:
@kyleam Another update that will probably won't be straightforward to process; it wan't obvious to me what the right behavior should be (until I consulted NONMEM); but I think at minimum we were inconsistent in how EVID==4 with SS==1 ws handled with and without lag time. 

There are benchmarks against NONMEM as well as demonstration of equivalent outputs to establish what we're claiming for behavior. 

Thanks in advance for looking at this.

mrgsolve.Rproj
```diff
@@ -1,4 +1,5 @@
 Version: 1.0
+ProjectId: 0063b92e-a050-4fd7-986e-be62f57eb1b1
```
kyleam, line 2: Just a heads up that this will conflict with the project ID from gh-1266.

kyleam (approved):
Thanks for the extended context and background as well as the nmtests links.

mrgsolve.Rproj
kylebaron: Thanks, @kyleam

Issue #1237, opened 2024-09-23
Error when trying to update with single non-updatable item


kylebaron:
Can't believe we never ran into this before: 

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '
$SET a = 2
'
mod <- mcode("foo", code, compile = FALSE)
#> Error in a[valid] <- all_updatable[m[valid]]: replacement has length zero
```

<sup>Created on 2024-09-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

But this works

``` r
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '
$SET 
a = 2
end = 24
'
mod <- mcode("foo", code, compile = FALSE)
```

<sup>Created on 2024-09-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This only happens when you update via `data` argument:
```r
> mod <- house() %>% update(data = list(a = 2))
Error in a[valid] <- all_updatable[m[valid]] : 
  replacement has length zero
```

PR #1271, opened 2025-03-15
Validate names supplied to $SET

kylebaron:
See #1237 

Basically, the `$SET` block can take some items that get passed directly to `update()` and other items that get used for various things under the hood. The problem reported in #1237 shows what can happen when an invalid name gets passed in (beyond just not getting picked up and notification sent to user). 

This PR centralizes the names of all items that can be in play  via `$SET` and `update()` and implements a `handle_()` function to validate inputs, generating an error if an input is bad.

### DIFF
--- a/R/compile.R
+++ b/R/compile.R
@@ -25,7 +25,6 @@ generate_rdefs <- function(pars,
                            model = "",
                            omats,
                            smats,
-                           set = list(), 
                            plugin = NULL,
                            dbsyms = FALSE, ...) {
 
@@ -73,7 +72,7 @@ generate_rdefs <- function(pars,
 
     Fdef <- Adef <- Ddef <- Rdef <- cmtndef <- NULL
 
-    cmtn <- unique(intersect(cvec_cs(set$CMTN),cmt))
+    cmtn <- unique(intersect(cvec_cs(cmtn,cmt))
     
     # Handle plugins; note plugin[["nm-defs"]] are punched in below     
     if(!is.null(plugin[["N_CMT"]])) cmtn <- cmt  
--- a/R/compile.R
+++ b/R/compile.R
@@ -25,6 +25,7 @@ generate_rdefs <- function(pars,
                            model = "",
                            omats,
                            smats,
+                           cmtn = NULL, 
                            plugin = NULL,
                            dbsyms = FALSE, ...) {
 
@@ -72,7 +73,7 @@ generate_rdefs <- function(pars,
 
     Fdef <- Adef <- Ddef <- Rdef <- cmtndef <- NULL
 
-    cmtn <- unique(intersect(cvec_cs(cmtn,cmt))
+    cmtn <- unique(intersect(cvec_cs(cmtn),cmt))
     
     # Handle plugins; note plugin[["nm-defs"]] are punched in below     
     if(!is.null(plugin[["N_CMT"]])) cmtn <- cmt  
--- a/R/handle_spec_block.R
+++ b/R/handle_spec_block.R
@@ -994,7 +994,7 @@ handle_SET <- function(spec) {
   ans <- tolist(dump_opts(spec[["SET"]]))
   valid <- c(GLOBALS$UPDATE_ALL, GLOBALS$SET_EXTRA, GLOBALS$SET_ARGS)
   incoming <- charmatch(names(ans), valid)
-  bad <- which(is.na(incoming))
+  bad <- which(is.na(incoming) | incoming == 0)
   if(!length(bad)) {
     return(ans)  
   }
--- a/R/handle_spec_block.R
+++ b/R/handle_spec_block.R
@@ -993,15 +993,17 @@ BLOCK <- function(x, env, type, code = NULL, get = NULL, ...) {
 handle_SET <- function(spec) {
   ans <- tolist(dump_opts(spec[["SET"]]))
   valid <- c(GLOBALS$UPDATE_ALL, GLOBALS$SET_EXTRA, GLOBALS$SET_ARGS)
-  bad <- setdiff(names(ans), valid)
-  if(length(bad)) {
-    if(length(bad) > 1) { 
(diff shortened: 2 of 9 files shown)


R/compile.R
```diff
@@ -25,7 +25,7 @@ generate_rdefs <- function(pars,
                            model = "",
                            omats,
                            smats,
-                           set = list(), 
+                           cmtn = NULL,
```
kylebaron, line 28: Just pass in what we need here.

R/mread.R
```diff
@@ -289,11 +289,6 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
   mread.env[["audit_dadt"]] <- 
     isTRUE(audit) && isTRUE(mread.env[["audit_dadt"]])
   
-  # Look for compartments we're dosing into: F/ALAG/D/R
```
kylebaron, line 292: Just relocating this

R/update.R
```diff
@@ -17,15 +17,6 @@
...
 
-sval <- unique(c("atol","rtol","ss_rtol", "ss_atol",
-                 "verbose","debug","preclean","mindt",
-                 "digits", "ixpr", "mxhnil","start", "end", "add", "delta",
-                 "maxsteps", "hmin", "hmax","tscale", "request"))
-
-other_val <- c("param", "init", "omega", "sigma", "outvars")
-
-all_updatable <- c(sval,other_val)
-
```
kylebaron, line 28: These are all replaced with vectors in `GLOBALS`

R/handle_spec_block.R
```diff
@@ -989,3 +989,21 @@ BLOCK <- function(x, env, type, code = NULL, get = NULL, ...) {
...
+  }
+  if(length(bad) > 1) { 
+    msg <- "The $SET block cannot handle these items:"
+  } else {
+    msg <- "The $SET block cannot handle this item"
+  }
+  bad <- names(ans)[bad]
+  names(bad) <- rep("x", length(bad))
+  abort(msg, body = bad, call = caller_env())
+}
```
kylebaron, line 1009: When processing the `$SET` block, check the names of items passed in.

R/mread.R
```diff
@@ -200,7 +200,7 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
   
   ## Pull out the settings and ENV now
   ## We might be passing parse settings in here ...
-  SET <- tolist(dump_opts(spec[["SET"]]))
+  SET <- handle_SET(spec)
```
kylebaron, line 203: `$SET` needs special processing, before processing all the other blocks.

R/handle_spec_block.R
```diff
@@ -989,3 +989,21 @@ BLOCK <- function(x, env, type, code = NULL, get = NULL, ...) {
   x <- structure(.Data=code, class=paste0("spec",type),pos=1)
   handle_spec_block(x,env)
 }
+
+handle_SET <- function(spec) {
+  ans <- tolist(dump_opts(spec[["SET"]]))
+  valid <- c(GLOBALS$UPDATE_ALL, GLOBALS$SET_EXTRA, GLOBALS$SET_ARGS)
+  incoming <- charmatch(names(ans), valid)
+  bad <- which(is.na(incoming))
```
kyleam, line 997: This doesn't catch the original case reported in gh-1237:

```
$ git rev-parse HEAD
e4a529b3eea4f3eac51902dcac8217875c9f1f9d
```

```R
library(mrgsolve)

code <- '
$SET a = 2
'
mod <- mcode("foo", code, compile = FALSE)
#> Error in a[valid] <- GLOBALS$UPDATE_ALL[m[valid]] :
#>   replacement has length zero
```

In that case, `charmatch` returns 0 due to the ambiguous partial match.

How about something like this on top?

```diff
diff --git a/R/handle_spec_block.R b/R/handle_spec_block.R
index 977f6b85..7d6d2407 100644
--- a/R/handle_spec_block.R
+++ b/R/handle_spec_block.R
@@ -994,7 +994,7 @@ handle_SET <- function(spec) {
   ans <- tolist(dump_opts(spec[["SET"]]))
   valid <- c(GLOBALS$UPDATE_ALL, GLOBALS$SET_EXTRA, GLOBALS$SET_ARGS)
   incoming <- charmatch(names(ans), valid)
-  bad <- which(is.na(incoming))
+  bad <- which(is.na(incoming) | incoming == 0)
   if(!length(bad)) {
     return(ans)  
   }
diff --git a/R/update.R b/R/update.R
index cbe54599..7eb3a5bb 100644
--- a/R/update.R
+++ b/R/update.R
@@ -121,8 +121,8 @@
     msg <- c(msg, bad)
     warn(msg, call = NULL)
   }
-  
-  valid <- !is.na(m)
+
+  valid <- !is.na(m) & m != 0
   a[valid] <- GLOBALS$UPDATE_ALL[m[valid]]
   names(args) <- a
   valid.in <- which(a %in% GLOBALS$UPDATE_SINGLE)
diff --git a/tests/testthat/test-modspec.R b/tests/testthat/test-modspec.R
index 4719009f..139be838 100644
--- a/tests/testthat/test-modspec.R
+++ b/tests/testthat/test-modspec.R
@@ -767,6 +767,12 @@ test_that("Invalid item in $SET generates error", {
     "The $SET block cannot handle this item", 
     fixed = TRUE
   )
+  code <- "$SET a = 2"  # ambiguous partial match
+  expect_error(
+    mcode("dollar-set-item-check-1", code),
+    "The $SET block cannot handle this item",
+    fixed = TRUE
+  )
   code <- "$SET end = 25, kyle = 2, zip = 55455"
   expect_error(
     mcode("dollar-set-item-check-2", code),
```

kylebaron: Thanks; all suggestions adopted into the PR. I re-discovered the partial matching there while working on this and put it on the list of things to get rid of.

kyleam (approved):

Issue #1275, opened 2025-03-29
Refactor nrow for matlist objects


kylebaron:
This will currently break when there is an empty matlist object. 

It should look at the `n` slot rather than calling `nrow()` on `x@data`.

PR #1276, opened 2025-03-29
refactor c method for matlist to return all non-skipped matrices

kylebaron:
This pull request fixes a bug when a model imports nonmem estimates from multiple runs, but doesn't request all the omega matrices (#1274). The issue was faulty bookkeeping in the `c()` function. 

I opened another issue #1275 to refactor some functionality around determining the dimensions of these objects; right now, I just do it brute force and it works fine; but some of that functionality (1) could be used here and (2) could break depending on whether or not the matlist has "empty" slots.


kylebaron:
Thank you!

PR #1282, opened 2025-04-18
Clobber both upper and lower case names when writing model

kylebaron:
Force block names to be upper case when processing and on write. So, when writing a model back to file, it will _always_ use `$BLOCK` style.  This will be a change, but it's reasonable.

### DIFF
--- a/R/mwrite.R
+++ b/R/mwrite.R
@@ -120,6 +120,9 @@ mwrite_model_to_list <- function(x) {
   code <- modelparse(code, comment_re = character(0))
   code <- lapply(code, trimws, which = "right")
   
+  names(code) <- toupper(names(code))
+  code_names <- names(code)
+  
   l$set <- tolist(code$SET)
   
   if(nrow(x@annot$data)) {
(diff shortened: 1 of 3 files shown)


kyleam (approved):
Makes sense to me.

Left a non-blocking suggestion for a test.

R/mwrite.R
```diff
@@ -120,6 +120,8 @@ mwrite_model_to_list <- function(x) {
   code <- modelparse(code, comment_re = character(0))
   code <- lapply(code, trimws, which = "right")
   
+  names(code) <- toupper(names(code))
```
kyleam, line 123: Is it worth adding something like below as a test?

```R
test_that("mwrite normalizes block names to upper case", {
  skip_if_not_installed("yaml")

  ucode <- '$CMT DEPOT CENT
$PKMODEL ncmt=1, depot=TRUE
$MAIN
double CL = 1
double V = 20
double KA = 1'
  umod <- mcode("test-case", ucode, compile = FALSE)
  utemp <- tempfile()
  mwrite_yaml(umod, file = utemp)

  lcode <- '$cmt DEPOT CENT
$pkmodel ncmt=1, depot=TRUE
$main
double CL = 1
double V = 20
double KA = 1'
  lmod <- mcode("test-case", lcode, compile = FALSE)
  ltemp <- tempfile()
  mwrite_yaml(lmod, file = ltemp)

  expect_identical(readLines(utemp), readLines(ltemp))
})
```

kylebaron: Thanks @kyleam; I had a simple test in place, but looks like I didn't push that. Will implement the proposed test and make sure everything is pushed.

tests/testthat/test-mwrite.R
```diff
@@ -230,3 +230,43 @@ test_that("code gets appropriately quoted", {
   lines <- readLines(cpp)
   expect_match(lines, 'outvars = "A"', all=FALSE)
 })
+
+test_that("model with lower case names", {
```
kyleam, line 234: missing `skip_if_not_installed("yaml")`

new test looks good to me otherwise

kylebaron: Thanks.

PR #1287, opened 2025-07-04
Fix floating point issue with time in some output records 

kylebaron:
We've see several reports in the past year where `TIME` in simulated output had some very small inaccuracies, clearly related to floating point arithmetic. 

This only happened when 
1. There was a random lag time
2. There were additional doses
3. The lag time made the scheduled "parent" dose the time of another record
4. Only observation records were requested in the output

I tracked this down to the code involved in this PR; basically, in upstream code we're adding a lag time to the time of the parent dose; then in the `schedule()` function, we're repeatedly subtracting that lag time off in order to put "parent" doses into the record stack. 

This is clearly where the issue is happening, but I still don't know _why_ it's happening. All the affected records should be doses. 

In this PR, we calculate the time of the parent dose once, rather than repeatedly. This seems to fix the issue.

### DIFF
--- a/src/datarecord.cpp
+++ b/src/datarecord.cpp
@@ -555,14 +555,17 @@ void datarecord::schedule(reclist& thisi, double maxtime,
     this_evid = Rate > 0 ? 5 : 1;
   }
   
-  //thisi.reserve(thisi.size() + n_dose); // TODO: remove 
-  
   double ontime = 0;
+  
+  // Doing this math once fixes issues with very small differences in output 
+  // TIME values: 
+  // https://github.com/metrumresearchgroup/mrgsolve/issues/1286
+  double parent_time = Time - lagt;
   
   int mp = 1000000000;
   
   int nextpos = addl_ev_first ?  -1000000000 : mp;
-  
+
   for(unsigned int k = 1; k <= Addl; ++k) {
     
     ontime = Time + Ii*double(k);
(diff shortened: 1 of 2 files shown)


src/datarecord.cpp
```diff
@@ -555,24 +555,29 @@ void datarecord::schedule(reclist& thisi, double maxtime,
...
     
     if(add_parent_doses) {
-      rec_ptr ev_parent = NEWREC(Cmt, this_evid, Amt, ontime-lagt, Rate, nextpos, Id);
-      ev_parent -> unarm(); 
-      ev_parent -> phantom_rec();
+      rec_ptr ev_parent = NEWREC(Cmt, this_evid, Amt, 
+                                 parent_time + Ii*double(k), 
+                                 Rate, nextpos, Id);
+      ev_parent->unarm(); 
+      ev_parent->phantom_rec();
```
kylebaron, line 580: Remove spaces around `->`

kyleam (approved):
The change looks good to me (the new form is mathematically equivalent and its order should be numerically more stable).

> This is clearly where the issue is happening, but I still don't know _why_ it's happening.

Based on the test case, it happens in a subset of cases where `lagt` is identical to `Time`.  The difference in the new and old code comes down to order of operations.  The old code adds a value to `Time` and then subtracts `lagt`.  The new code subtracts `lagt` from `Time` and then adds the other value.

```
old: Time + Ii*double(k) - lagt
new: Time - lagt + Ii*double(k)
```

---

Here are input values from a case where the `old` and `new` values do not align:

```
  Time:     8.4836259969106730949306438560597598552703857421875
  lagt:     8.4836259969106730949306438560597598552703857421875
    Ii:    12.0000000000000000000000000000000000000000000000000
     k:     1.0000000000000000000000000000000000000000000000000
```

Different values for `old` and `new` order:

```R
x <- 8.4836259969106730949306438560597598552703857421875
sprintf("%.49f", x + 12 - x)
#> [1] "12.0000000000000017763568394002504646778106689453125"
sprintf("%.49f", x - x + 12)
#> [1] "12.0000000000000000000000000000000000000000000000000"
```

kylebaron:
Oh cool ... thanks for tearing that apart, @kyleam.

curtisKJ:
Thanks for sorting this out.

PR #1291, opened 2025-07-11
Continuous monitoring

kylebaron:
This PR adds code allowing users to insert non-output observation or other-type event records into the simulation from inside the model. Non-output means the records will not appear in the simulated output. These records are only for internal monitoring of the system to make decisions (e.g. for dynamic dosing).

New `evtools` function: `evt::tgrid(<start>, <end>, <delta>)`
New `evtools` method: `evt::evid(<event object>, <new evid>)`

New behavior: events that were started in the past were silently ignored; this seems very reckless and I had to deal with it when thinking through whether to let users specify a time grid starting at `time = 0` but actually implementing that _after_  that (`time > 0`). It seems so like we just need to stop at that point and ask the user to start it `now` at the earliest. 

New behavior: `EVID`s that are 100 or greater will not introduce a discontinuity in the simulation; those records will behave like observation records with respect to the solver (no discontinuity), but will continue to carry all the requirements of dosing records. 

The default `EVID` for the records added by `evt::tgrid()` have `EVID==0`. This is the most intuitive result, but probably not the one that people should want most of the time for something like this. On the fence right now if I should default to something like 999 instead and let people opt back to `0`? or default to `0` and let people opt to 999.

### DIFF
--- a/inst/maintenance/unit-cpp/test-evtools-tgrid.R
+++ b/inst/maintenance/unit-cpp/test-evtools-tgrid.R
@@ -40,15 +40,16 @@ test_that("Insert an internal-only time grid with NEWIND", {
   n <- length(seq(mod$xstart, mod$xend, mod$xdelta))
   mod <- update(mod, end = -1, add = c(0, mod$xend + 3))
   
+  # Count all EVID=0 records
   mod <- param(mod, evi = 0, mode = 0)
   out <- mrgsim(mod, recsort = 4)
   expect_equal(nrow(out), 2)
-  expect(out$count[2], n)
+  expect_equal(out$count[2], n + length(stime(mod)))
   
   mod <- param(mod, evi = 999, mode = 0)
   out2 <- mrgsim(mod, recsort = 4)
   expect_equal(nrow(out2), 2)
-  expect(out2$count[2], n)
+  expect_equal(out2$count[2], n)
   
   mod <- param(mod, evi = 2, mode = 0)
   out3 <- mrgsim(mod)
@@ -64,7 +65,7 @@ test_that("Insert internal-only time grid with odd delta", {
   mod <- param(mod, evi = 999, mode = 0)
   out <- mrgsim(mod, recsort = 4)
   expect_equal(nrow(out), 2)
-  expect(out$count[2], n)
+  expect_equal(out$count[2], n)
 })
 
 test_that("Insert internal-only time grid starting at other time", {


src/datarecord.cpp
```diff
@@ -206,10 +206,10 @@ void datarecord::implement(odeproblem* prob) {
     } else {
       this->evid(1);
     }
-    this-> implement(prob);
+    this->implement(prob);
```
kylebaron, line 209: Just removes a space.

src/datarecord.cpp
```diff
@@ -206,10 +206,10 @@ void datarecord::implement(odeproblem* prob) {
...
     return;
   }
-  prob->lsoda_init();
+  if(Evid < 100) prob->lsoda_init();
 }
```
kylebaron, line 213: New behavior: don't reset the solver if evid is 100 or greater

src/devtran.cpp
```diff
@@ -633,9 +633,21 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
...
+          if(this_time < tto && !mt[mti].now) {
+            throw Rcpp::exception(
+                tfm::format(
+                  "modeled events or observations cannot start in the past\n"
+                  "ID: %d, time: %d, modeled record time: %d, evid: %i", 
+                  id, tto, this_time, this_evid
+                ).c_str(),
+                false
+            );
+          };
```
kylebaron, line 646: New behavior: if the event _starts_ in the past, we throw an error

inst/maintenance/unit-cpp/test-evtools-dosing.R
```diff
@@ -46,8 +50,12 @@ if(TIME==0 && Mdose > 0) {
   if(Cmt2 > 0) evt::cmt(dose, Cmt2);
   if(Amt2 > 0) evt::amt(dose, Amt2);
   if(Time2 > 0) evt::retime(dose, Time2);
+  if(Evid2 > 0) evt::evid(dose, Evid2);
   self.push(dose);
 }
+cumevid = cumevid + EVID;
+
+$CAPTURE cumevid
 '
```
kylebaron, line 59: Tweaking this test model to allow us to tickle `evt::evid()`.

inst/base/mrgsolve-evtools.h
```diff
@@ -116,11 +116,27 @@ void cmt(mrgsolve::evdata& ev, const int cmt) {
   return;
 }
 
+void evid(mrgsolve::evdata& ev, const unsigned int evid) {
+  ev.evid = evid;
+  return;
+}
+
 void now(mrgsolve::evdata& ev) {
```
kylebaron, line 124: New method: set `evid` attribute.

src/datarecord.cpp
```diff
@@ -588,6 +588,24 @@ void datarecord::schedule(reclist& thisi, double maxtime,
   }
 }  
 
+/*  Insert observation records triggered from evtools
+ * 
+ */
+
+void insert_observations(reclist& thisi, mrgsolve::evdata& ev, const size_t start, 
+                         const bool put_ev_first) {
+  const int total = ev.addl + 1;
```
kylebaron, line 597: Notice that we're incrementing `addl` here, to insert the equivalent number of records as we would for a dose.

src/datarecord.cpp
```diff
@@ -588,6 +588,24 @@ void datarecord::schedule(reclist& thisi, double maxtime,
...
+  int nextpos = put_ev_first && (ev.evid!= 1 && ev.evid!=1 && ev.evid!=4) ? -1000000000 : 1000000000;
```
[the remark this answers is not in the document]
kylebaron: De-duplicated.

inst/maintenance/unit-cpp/test-evtools-tgrid.R
```diff
@@ -0,0 +1,87 @@
...
+
+test_that("Insert an internal-only time grid with NEWIND", {
+  
+  n <- length(seq(mod$xstart, mod$xend, mod$xdelta))
+  mod <- update(mod, end = -1, add = c(0, mod$xend + 3))
+  
+  mod <- param(mod, evi = 0, mode = 0)
+  out <- mrgsim(mod, recsort = 4)
+  expect_equal(nrow(out), 2)
+  expect(out$count[2], n)
```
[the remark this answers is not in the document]
kylebaron: Thanks; updating that test. It counts all `EVID=0` records, so adding the number of output rows into the count.

src/datarecord.cpp
```diff
@@ -588,6 +588,24 @@ void datarecord::schedule(reclist& thisi, double maxtime,
...
 }  
 
+/*  Insert observation records triggered from evtools
+ * 
+ */
+
+void insert_observations(reclist& thisi, mrgsolve::evdata& ev, const size_t start, 
+                         const bool put_ev_first) {
+  const int total = ev.addl + 1;
+  int nextpos = put_ev_first && (ev.evid!= 1 && ev.evid!=4) ? -1000000000 : 1000000000;
```
kyleam, line 598: I see that -1000000000 and 1000000000 align with the handling of `addl_ev_first` in `datarecord::schedule`. Okay.

kyleam (approved):
Thanks, as usual, for the detailed PR description. Some of the details of datarecord.cpp/devtran.cpp are beyond me, but the new code looks fine to me and aligns with what I expect based on reviewing surrounding code.

For documentation of trig/evid, I'm guessing that will come as updates for <https://mrgsolve.org/user-guide/plugins.html#sec-plugin-evtools>?

kylebaron:
>For documentation of trig/evid, I'm guessing that will come as updates for https://mrgsolve.org/user-guide/plugins.html#sec-plugin-evtools?

Yeah, I'll have to document there once this is released. I'll also document the new EVID behavior. I don't think I really said anything about what happens when you, e.g., pass in `EVID=147`. I'll probably spell that out explicitly for people. Also on the radar is to partition off a part of the EVID space for reserved EVIDs (like stopping an infusion); but I'd want to be very careful about that.

PR #1294, opened 2025-07-22
Exact matching when aliasing block names

kylebaron:
I think there was an issue when mistyping a block name like this:

```
$ODEFOO
```

Which gets turned into 

```
$DESFOO
```

The block name aliasing should be / was always intended to be exact matching on the whole name. I'm not sure of a another one-liner in base R to do this or not.


kylebaron:
Thanks for the test, @kyleam. Looks good to me.

PR #1307, opened 2025-11-23
Change error to warning for check_modeled_infusions

kylebaron:
PR #1304 implemented a check of the data set `RATE` for dosing records when the infusion rate was modeled in `$MAIN`. The check resulted in an error if the data set `RATE` wasn't as expected (-1 for modeled rate and -2 for modeled duration). 

This PR demotes the error to a warning. Some refactoring was needed to avoid a warning when checking the data rate against modeled rate or duration getting issued right before an error when checking the other direction.

### DIFF
--- a/inst/maintenance/unit/test-D-R-F.R
+++ b/inst/maintenance/unit/test-D-R-F.R
@@ -68,7 +68,8 @@ data$rate <- -2
 test_that("Error when rate = -2 and R_CMT set instead of D_CMT", {
   expect_error(
     mod %>% data_set(data) %>% mrgsim(), 
-    "invalid infusion duration"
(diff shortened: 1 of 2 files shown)


kyleam (approved):
Just to capture a little of the offline discussion about this (@kylebaron, of course please correct me if any of this is off): the check signaling an error works well in terms of the desired behavior for end users.  However, the check trips up a campsis test, raising the concern that the new behavior is too rigid in the context of tools building on top of mrgsolve.  Demoting the error to a warning gives developers of downstream tools more flexibility to adjust things as they see fit (adding `@!check_modeled_infusions` or swallowing the warning or tweaking their approach to avoid the warning entirely).

---

The changes look good to me.  I've confirmed that the campsis test passes, as expected.

PR #1310, opened 2026-01-07
Write code to file with newline separator

kylebaron:
I think the original implementation erroneously didn't sep by `\n`.  With this update, I can pass a character vector (length > 1) of `code` in to `mcode()` and compile it. I want to start doing this:

```r
old_mod <- mread("foo/bar.mod")

new_mod <- mcode("new-model", code = old_mod@code)
```


kyleam (approved):
> I want to start doing this:

Makes sense.

Looks good to me (though I think it'd be worth adding a simple test of the behavior).

PR #1314, opened 2026-01-14
Release/1.7.2

kylebaron:
# mrgsolve 1.7.2

- An individual record counter has been added. Use `self.inrow` to get the number of
  output rows for the current individual; use `self.irown` to get the zero-indexed row
  number for the current individual; these are similar to `self.nrow` and `self.rown`, 
  which count records for the entire output data set (#1313). 

## Bugs Fixed

- The `code` argument to `mcode()` now accepts a character vector with length 
  greater than 1; passing multiple lines previously caused a compile error (#1310).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # mrgsolve 1.7.2
 
-- And individual record counter has been added. Use `self.inrow` to get the number of
+- An individual record counter has been added. Use `self.inrow` to get the number of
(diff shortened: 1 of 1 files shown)


kylebaron:
## revdepcheck results

We checked 6 reverse dependencies, comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 0 new problems
 * We failed to check 0 packages

PR #1327, opened 2026-02-13
Macros to determine the last record for an individual or a simulation

kylebaron:
- `FINAL_ROW` signals we just advanced to the final row of the output data
- `FINAL_IROW` signals we just advanced to the final row of the current individual
 
Both have been added to the RESERVED word list. 

See #1328

### DIFF
--- a/src/devtran.cpp
+++ b/src/devtran.cpp
@@ -461,16 +461,17 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
         Rcpp::checkUserInterrupt();
         ic = prob.interrupt;
       }
-      
+    
       rec_ptr this_rec = a[i][j];
       
       this_rec->id(id);
       tto = this_rec->time();
       
+      // TODO: simplify
       if(icrow==NNI || crow==NN || tto > maxtime) {
         continue;
       }
-    
+      
       // Only update row counters on output records
       if(this_rec->output()) {
          prob.irown(icrow);


src/devtran.cpp
```diff
@@ -460,17 +462,20 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
...
+        continue;
+      }
+    
       // Only update row counters on output records
       if(this_rec->output()) {
          prob.irown(icrow);
          prob.rown(crow);
-      }      
-
-      this_rec->id(id);
```
kylebaron: Moved this up for clarity.

inst/maintenance/unit/test-rown-nrow.R
```diff
@@ -263,7 +277,47 @@ test_that("row counters are only updated on output records gh-1323", {
   expect_equal(out$total1[6], 1)
   expect_equal(out$total2[6], 1);
   
+  # Macros
+  expect_equal(out$total1, out$total3)
+  expect_equal(out$total4, out$total4)
+})
+
+test_that("more row counter macros gh-132X", {
```
kyleam: Placeholder that should be updated to `gh-1327`?

src/devtran.cpp
```diff
@@ -460,17 +462,20 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
...
       
-      if(crow == NN) continue;
-      
       rec_ptr this_rec = a[i][j];
-
+      
+      this_rec->id(id);
+      tto = this_rec->time();
+      
+      if(icrow==NNI || crow==NN || tto > maxtime) {
```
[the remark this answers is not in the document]
kyleam: Thanks for the details/explanation.  Leaning conservative/safe is of course fine by me.

> Could you test with just `icrow==NNI`?

Yes, the tests still all pass on my end with the condition reduced to `if(icrow==NNI)`.

PR #1355, opened 2026-03-26
Tweak env_get to provide access to objects in envir

kylebaron:
@kyleam please look and suggest. I mainly want a function to `get()` objects from the model environment. Since the model object gets stored in the output object, we can go to either. There was some existing stuff, but I didn't like the way it was set up. So I tried to adapt as best I could while not straying too far from what was in place. But TBH, there's probably not a lot of people actually using this `env_*() series. 

<hr>

1. `env_get()` is now generic; methods for `mrgmod` and`mrgsims`
2. Add argument `what` that is passed to `base::get()`; note this is a new / different argument in second position
3. Adds some tests

Known that `env_get()` is also exported now by `rlang`. Considered providing a `.default` method that would dispatch there if someone passed an environment. But cut back for simplicity. Thoughts?  I'm not _super_ concerned about conflicts, but if something simple could be done to provide a nicer experience, I'd do it.


kyleam (approved):
Reviewed latest changes and didn't spot any issues.  As discussed, you're okay with the breaking change to `env_get`, so that works for me.

My vote would be to drop the `env_get_obj` alias, but either way is fine.

PR #1360, opened 2026-03-30
Preprocessing C++ code 

kylebaron:
# DSL Preprocessing

This PR introduces a C++-based preprocessing layer (`src/dsl-preprocess.cpp`)
that transforms NONMEM-style model code into valid C++ that can be used in 
mrgsolve. This does not change the users source code; rather, modifications 
are made under the hood and only appear in the `.cpp` file that is actually
compiled; these changes are on top of a host of changes that mrgsolve is already 
doing. The difference is that the current work is using compiled Boost 
functionality rather thand regex.

## History
I've been wanting to do all of this for a while. I did some iteration trying
this out with R functions, but it became complex quick. 

I asked claude to write and R function to `convert_pow()` and it did it. I
pressed into some edge cases and asked about performance and it suggested the Qi
stuff in Boost (already in mrgsolve production code). We started going down this
road, hoping we could accomplish other objectives (like `warn_int_div()`) and so
far worked well.

claude also started a lot of the tests. We did a lot of iteration as I got into
the tests as well, thinking through different patters users are likely to use in
their code and making sure the parser behaved the same as NONMEM. I wrote 100%
of the integration into the package (how these low level processor functions are
used when processing a model file), including the plugins and the environment
variables. 

## Exponents: `**` to  `pow()` conversion

The first piece was converting Fortran-style exponentiation (`**`) to C++
`pow()` calls. The implementation uses a Boost.Spirit parser to build a full
expression AST rather than doing regex substitution, so operator precedence and
parenthesization are handled correctly. Initial prototyping used a series of R
functions to do this; it didn't work that bad, but I could see that there would
be edge cases that would be difficult to handle this this approach.

A warning is issued if a line contains `**` but the parser cannot convert it
(e.g., due to an unrecognized syntax), so the user is told rather than silently
left with bad code. The original line is passed through unchanged when
conversion fails.

This runs by default for `$MAIN`, `$TABLE`, `$ODE`, `$EVENT`, `$PREAMBLE`, and
`$PRED` blocks (see `GLOBALS$PRE_PROC_BLOCKS`). It can be turned off via an
environment variable (`MRGSOLVE_CONVERT_POW = FALSE`) in `$ENV`.

While it is possible to have `**` in C++ code, it should be very rare; so I feel
good about processing this by default. If the user never has `**` in their code,
this will do nothing. 

There is an exported function called `convert_pow()` that will process a
character vector of code.

## Integer-division warning

Building on the same AST infrastructure, the preprocessor walks parsed
expressions looking for division of two bare integer literals (e.g. `3/4`,
`1/2`). In C++, these truncate toward zero, so `3/4` evaluates to `0`. This is a
big problem for people coming from NONMEN world, which handles `3/4` as `0.75`;
we've had senveral complaints of users getting bit by this, despite warnings in
the user guide. A warning is issued for each occurrence, with a suggestion to
use `3.0/4.0` instead. 

This also runs by default across the same blocks and can be disabled with
`MRGSOLVE_WARN_INT_DIV = FALSE` in `$ENV`.

I think it's the right decision to warn by default on this; can't imagine where
you would really need integer `3/4` and there are ways to accomplish this in the
code or just set the escape env var to disable.

There is an exported function called `warn_int_div()` that will process a
character vector of code. 

## Fortran IF/THEN/ELSE/ENDIF conversion

Fortran block-structured conditionals and single-line `IF` forms are converted
to C++. For example:

```
IF (WT .GE. 90) THEN
  CL = CL * 0.8
ENDIF
```

becomes:

```cpp
if(WT >= 90) {
  CL = CL * 0.8;
}
```

This handles `ELSEIF`/`ELSE IF`, single-line `IF(cond) statement` forms, bare
`ELSE` and `ENDIF`/`END IF`, and converts all Fortran relational and logical
operators (`.GE.`, `.LE.`, `.GT.`, `.LT.`, `.EQ.`, `.NE.`, `.AND.`, `.OR.`,
`.NOT.`, `.TRUE.`, `.FALSE.`) case- insensitively. The conversion runs only when
the `nm-vars` plugin is active (see below), and can be disabled via
`MRGSOLVE_CONVERT_FORT_IF = FALSE` in `$ENV`.

Like the `convert_pow()` functionality, if the user never uses these constructs,
the processor should just pass the code through. This also will never touch code
that doesn't invoke `nm-vars` plugin. The environment variable escape route is
provided just in case we have some false positive detection of code to modify.

There is an exported function called `convert_fort_if()` that will process a
character vector of code.

## Semicolon insertion

A semicolon inserter adds trailing semicolons to statement lines that lack them,
to support NONMEM-style code where semicolons are optional. Lines ending with a
continuation operator (`+`, `-`, `*`, `/`, `&`, `,`, `=`, `(`) are left alone —
this covers multi-line expressions. Lines that are blank, brace-only,
preprocessor directives, or C/C++ comments are also left untouched.

An important edge case: Fortran uses `;` as an inline comment delimiter. If a
`;` is found mid-line and nothing after it looks like C++ (no further `=` or
`;`), the text after the semicolon is stripped as a Fortran comment. This
preserves compatibility when someone has Fortran-style inline comments in their
code.

Semicolon insertion is off by default. It is enabled by adding the `semicolons`
plugin (or the composite `nm-like` plugin) to `$PLUGIN`. It only operates in
combination with `nm-vars`.

This will be very helpful if it works right. And it should be fine as long as
someone is truly coming from NONMEM. This is why you not only have to be using
`nm-vars` but you also need to opt in through the `semicolons` plugin.

There is an exported function `convert_semicolons()` that will process a
character vector of code.

## Plugin integration: `nm-like` composite plugin

A new `nm-like` composite plugin was added as a convenience: invoking `$PLUGIN
nm-like` automatically pulls in `nm-vars`, `autodec`, and `semicolons`. This
makes it easier to write fully NONMEM-style model code without individually
listing the component plugins. 

I think making the `semicolons` available along with the `IF/ELSE` processing, 
we need to emphasize that, when you're in NONMEM mode, you really need to 
commit to those patterns. Veering off into more complex C++ stuff could 
possibly not end well. I'm fine with this; IMO, invoking `nm-vars` is the 
user opting into this restriction; this can't be done lightly.
å
## User-facing R functions

The four preprocessing functions are exported and documented together under
`?dsl_preprocess`:

- `convert_pow(x, block = "")` — convert `**` to `pow()`
- `warn_int_div(x, block = "")` — warn about integer division
- `convert_fort_if(x)` — convert Fortran IF constructs
- `convert_semicolons(x)` — insert missing semicolons

These can be called directly on character vectors for testing or interactive
inspection.

The `block` argument is used when processing a model file; the warning or 
error includes the code in question (integer division or unparsed line) 
along with the block name where the code came from. We still can't report
line numbers, because of the way blocks are processed.

## Environment variable control

All three default-on preprocessors can be turned off from within a model file
using `$ENV`:

```r
$ENV
MRGSOLVE_CONVERT_POW      = FALSE
MRGSOLVE_CONVERT_FORT_IF  = FALSE
MRGSOLVE_WARN_INT_DIV     = FALSE
```

The flags are read from `$ENV` during `parse_env()` when `mread()` processes the
model, so they apply on a per-model basis rather than globally.


inst/maintenance/unit-cpp/test-cpp.R
```diff
@@ -18,6 +18,7 @@ DADT(1) = -0.1 * A(1);
 '
 
 test_that("build a model with mrgx and nm-vars", {
+  skip_if_not_installed("Rcpp")
```
kyleam: Why is this necessary when Rcpp is a hard dependency?

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+  std::string operator()(const ast::BinaryExpr& b) const {
+    std::string l = boost::apply_visitor(*this, b.lhs);
+    std::string r = boost::apply_visitor(*this, b.rhs);
+    if (b.op == "^") {
+      return "pow(" + l + ", " + r + ")";
+    }
+    if (needs_parens_left(b.lhs, b.op)) l = "(" + l + ")";
+    if (needs_parens_right(b.rhs, b.op)) r = "(" + r + ")";
+    return l + b.op + r;
```
kyleam, line 158: Based on the `convert_pow` tests, this seems like this "needs parens?" logic is working well.

Given this is behind-the-scenes translation, I'd be tempted to just always add the parens explicitly.

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+      _val = construct<ast::Number>(_1)
+    ];
+
+    // argument list for function calls
+    arglist = expr[push_back(_val, _1)]
+              >> *( lit(',') >> expr[push_back(_val, _1)] );
+
+    // function call: identifier followed immediately by '('
+    funccall = (identifier >> lit('(') >> arglist >> lit(')'))[
+      _val = construct<ast::FuncCall>(_1, _2)
```
kyleam, line 219: This may not be relevant in the context of NONMEM's abbreviated code and the C++ code used in mrgsolve, but it looks like it doesn't handle function calls without arguments.

```
convert_pow("x = fn()**2;")
#> [1] "x = fn()**2;"
#> Warning message:
#> Could not convert **: 'x = fn()**2;'
```

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+    // primary: number | function call | identifier | parenthesised expr
+    primary =
+        number                           [_val = _1]
+      | funccall                         [_val = _1]
+      | identifier                       [_val = _1]
+      | ( lit('(') >> expr >> lit(')') ) [_val = _1]
+      ;
+
+    // power: left-associative; base is primary so unary minus binds looser
+    // Fortran '**' stored as '^' in the AST; exponent may be unary (a**-2)
```
kyleam: This comment confuses me. It says "left-associative", but `**` is right-associative in Fortran (`2 ** 3 ** 2 => 2 ** (3 ** 2)`), and `convert_pow` handles it that way.

```
convert_pow("2 ** 3 ** 2")
#> "pow(2, pow(3, 2))"
```

> exponent may be unary (a\*\*-2)

Probably fine to represent, but NM-TRAN rejects this:

```
AN ERROR WAS FOUND ON LINE 21 AT THE APPROXIMATE POSITION NOTED:
 V = THETA(3)**-2
               X
 THE CHARACTERS IN ERROR ARE: -
  202  FORTRAN SYNTAX IS INCORRECT OR INAPPROPRIATE IN THIS CONTEXT.
```

Fortran gives a warning:

```
    5 |     res1 = 2**-4
      |               1
Warning: Extension: Unary operator following arithmetic operator (use parentheses) at (1)
 res     0.0000000000000000
```

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+    power =
+      primary[_val = _1]
+      >> *(
+        lit("**") >> unary[
+          _val = construct<ast::BinaryExpr>(_val, val(std::string("^")), _1)
```
kyleam: I think representing mapping power to `^` is unnecessarily confusing given that operator means something in C++ (even if it's not relevant in the context of mrgsolve). Can't the representation be anything (including, say, a more explicit "pow" or "power")?

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  qi::rule<Iterator, ast::FuncCall(),          asc::space_type> funccall;
+  qi::rule<Iterator, std::vector<ast::Expr>(), asc::space_type> arglist;
+  qi::rule<Iterator, std::string(),            asc::space_type> identifier;
+  qi::rule<Iterator, std::string(),            asc::space_type> addop;
+  qi::rule<Iterator, std::string(),            asc::space_type> mulop;
+  qi::rule<Iterator, std::string(),            asc::space_type> cmpop;
+};
+
+// ---------------------------------------------------------------------------
+// 3b. Integer-division detector
```
kyleam: There's no "3a".

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+// ---------------------------------------------------------------------------
+
+// Returns true if the Number text represents a plain integer (digits only,
+// no decimal point, no exponent suffix).
+static bool is_integer_literal(const ast::Number& n) {
+  const std::string& t = n.text;
+  return !t.empty() &&
+    t.find_first_not_of("0123456789") == std::string::npos;
```
kyleam, line 328: Should this consider negative numbers?

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  }
+  return i;
+}
+
+static std::string strip_line_comment(const std::string& s) {
+  auto pos = s.find("//");
+  return pos == std::string::npos ? s : s.substr(0, pos);
+}
+
+static std::string convert_line(const std::string& line) {
```
kyleam: `convert_line` as a name seems confusingly general given it's specific for the pow translation.

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  if (line.find("**") == std::string::npos) return line;
+
+  using It = std::string::const_iterator;
+  ExprGrammar<It> grammar;
+
+  auto eq = find_assign_eq(line);
+  std::string prefix, rhs_str;
+
+  std::string stripped_line = strip_line_comment(line);
```
kyleam, line 446: Both `find_assign_eq` and `strip_line_comment` look to just be characters scans, so I would think they'd get thrown off by not being aware of things like strings and all other sorts of C++ complexity.

But, those aren't relevant in the context of mrgsolve, right?

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+// Convert Fortran relational/logical operators to C++ equivalents.
+static std::string convert_fortran_ops(const std::string& s) {
+  std::string r = s;
+  r = replace_all_ci(r, ".GE.", ">=");
+  r = replace_all_ci(r, ".LE.", "<=");
+  r = replace_all_ci(r, ".GT.", ">");
+  r = replace_all_ci(r, ".LT.", "<");
+  r = replace_all_ci(r, ".EQN.", "==");
+  r = replace_all_ci(r, ".NEN.", "!=");
```
kyleam: `.EQN.` and `.NEN.` aren't Fortran operators as far as I can tell.

Does NM-TRAN support them? The NONMEM docs only talk about them in the context of the `$DATA` docs. I don't see mention of them in its abbreviated code docs. Under the "Conditional Statements" section, I see

> Conditions may include the operators .EQ., .NE., .LE., .GE., .LT., .GT., .AND., and .OR.. The arthmetic operators may be coded as `=, /`, <=, >=, <, and >, respectively.

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  r = replace_all_ci(r, ".EQ.", "==");
+  r = replace_all_ci(r, ".NE.", "!=");
```
kyleam, line 586: NM-TRAN/Fortran also support `/=` for "not equal". I confirmed that NM-TRAN generates the same FSUBS when either `.NE.` or `/=` is used. (NM-TRAN maps `.NE.` to `/=` underneath.)

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  r = replace_all_ci(r, ".AND.", "&&");
+  r = replace_all_ci(r, ".OR.", "||");
+  r = replace_all_ci(r, ".NOT.", "!");
+  r = replace_all_ci(r, ".TRUE.", "true");
+  r = replace_all_ci(r, ".FALSE.", "false");
```
kyleam: I don't think NM-TRAN supports `.NOT.`, `.TRUE.`, or `.FALSE.` for conditions in its abbreviated code.

```
AN ERROR WAS FOUND ON LINE 11 AT THE APPROXIMATE POSITION NOTED:
 IF(.NOT.FLAG) ET=1.3
             X
 THE CHARACTERS IN ERROR ARE: )
  202  FORTRAN SYNTAX IS INCORRECT OR INAPPROPRIATE IN THIS CONTEXT.
```

```
 AN ERROR WAS FOUND IN THE CONTROL STATEMENTS.

AN ERROR WAS FOUND ON LINE 13 AT THE APPROXIMATE POSITION NOTED:
 IF (.TRUE.) THEN
     X
 THE CHARACTERS IN ERROR ARE: .
   17  NUMERIC ITEM IS FORMED INCORRECTLY.
```

```
 AN ERROR WAS FOUND IN THE CONTROL STATEMENTS.

AN ERROR WAS FOUND ON LINE 13 AT THE APPROXIMATE POSITION NOTED:
 IF (.FALSE.) THEN
     X
 THE CHARACTERS IN ERROR ARE: .
   17  NUMERIC ITEM IS FORMED INCORRECTLY.
```

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+
+// Trim leading and trailing spaces/tabs.
+static std::string trim_ws(const std::string& s) {
+  size_t b = s.find_first_not_of(" \t");
+  if (b == std::string::npos) return "";
+  size_t e = s.find_last_not_of(" \t");
+  return s.substr(b, e - b + 1);
+}
+
+static std::string convert_fortran_if_line(const std::string& line) {
```
kyleam, line 612: I was surprised to see that this doesn't use any of the spirit/qi setup and just does plain character scanning. And, as a related point, I didn't expect to see it work at the line level (rather than all lines of the relevant mrgsolve block), where ops replacement can happen on any line and is not limited to if/else blocks.

I think pretty much arbitrary C++ code can occur in mrgsolve models, so it's of course not feasible to represent/parse that, but I was expecting to see some sort of partial representation of these Fortran-style blocks with a bunch of pass-throughs or something to accommodate C++.

But based on the tests, this line-scoped, limited-context character scanning and replacement is probably sufficient for mrgsolve's needs.

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+  // Preserve leading whitespace (indentation).
+  size_t iend = 0;
+  while (iend < line.size() &&
+         (line[iend] == ' ' || line[iend] == '\t')) ++iend;
+  const std::string indent = line.substr(0, iend);
+  const std::string body   = line.substr(iend);
+
+  if (body.empty()) return line;
+
+  // Skip C-style comments unchanged.
```
kyleam: This is per-line, and this will get confused by multi-line c-style comments. Again, probably not too relevant in the case of mrgsolve (and only consequence is doing substitution in a comment).

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+// Return true if s contains ch at paren depth 0.
+static bool contains_at_depth0(const std::string& s, char ch) {
+  int depth = 0;
+  for (size_t i = 0; i < s.size(); ++i) {
+    if      (s[i] == '(') ++depth;
+    else if (s[i] == ')') --depth;
+    else if (s[i] == ch && depth == 0) return true;
+  }
+  return false;
+}
```
kyleam: [ mostly thinking aloud, though I would be in favor of avoiding the duplication here ]

`find_semi_depth0` and `contains_at_depth0` have near duplicate logic. `find_semi_depth0` could be generalized to take a character and `contains_at_depth0` could call it underneath, converting the position return value to boolean. (Or `contains_at_depth0` could just be dropped, with the call sites handling the `std::strings::npos` check, which is a pattern that already happens a lot in this file.)

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+      (t[1] == '/' || t[1] == '*'))            return line;  // comment
+
+  // Only consider ';' at paren depth 0 — keeps for(;;) intact.
+  size_t semi = find_semi_depth0(line);
+
+  if (semi == std::string::npos) {
+    // No depth-0 ';': append one after the last non-whitespace character,
+    // unless the line ends with a continuation operator.
+    size_t last = line.find_last_not_of(" \t");
+    static const std::string cont_chars = "+-*/&,=(";
```
kyleam, line 766: Wouldn't the NM-TRAN `&` continuation operator fail at the C++ level?

And the rest for these characters seem only relevant for C++ code (NM-TRAN would fail on them). Ignoring that the character scan may get tripped up by comments, it seems hard to catch all possible C++ continuations. For example, I would think the set would also need `\` as a continuation character, but even then there are cases where there is no continuation character.

```
double result = (a * 2)
            / (b + 3);
```

Perhaps most of this isn't relevant in the context of mrgsolve. But I think it at least makes sense to make it clear that NM-TRAN's `&` will error and should not be used.

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+    if (cont_chars.find(t.back()) != std::string::npos) return line;
+    return line.substr(0, last + 1) + ";";
```
kyleam, line 768: Does this fail on lines ending with `//` comment?

```
convert_semicolons("a = b // comment")
#> [1] "a = b // comment;"
```

Also, I think it could mishandle valid C++ code like

```
if(IFLAG > 0)
  a = b;
```

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,867 @@
...
+        std::string prefix = block.empty()
+          ? "Integer division"
+          : "Integer division in $" + block + " block";
+        std::string msg =
+          prefix + ": '" +
+          inst.num + "/" + inst.den + "' truncates toward zero; " +
+          "use " + inst.num + ".0/" + inst.den + ".0 for real division";
+        warn_no_call(msg);
+      }
+    } catch (...) {}
```
kyleam, line 858: Is it intentional that this swallows all exceptions without any warning? The other spots have catch `std::exception` and warn. Perhaps all spots should catch `std::exception` and have a `...` catchall, both which warn that an unexpected exception was hit.

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  )
+})
+
+test_that("convert_pow handles nested **", {
+  expect_equal(
+    convert_pow("x = (a**2)**3;"),
+    "x = pow(pow(a, 2), 3);"
+  )
+  expect_equal(
+    convert_pow("x = 2**3**2;"),
```
kyleam, line 832: Good to see this case.

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  expect_equal(convert_pow("a**2"), "pow(a, 2)")
+  expect_equal(convert_pow("(WT/70)**0.75"), "pow(WT/70, 0.75)")
+})
+
+test_that("convert_pow handles c-style comment", {
+  code <- '
+  $GLOBAL 
+  /** 
+  This model is incredible.
+  */
```
kyleam, line 973: This case looks misleading. Inserting a `**` into that comment demonstrates that `convert_pow` doesn't ignore C++'s multi-line comments.

```
Warning message:
Could not convert **: '  This model is incredible. 3 ** 4
```

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  x <- lapply(lines, convert_pow)
+  x <- lapply(x, trimws)
+  expect_equal(x$MAIN, "double CL = THETA1*pow(WT/123, 0.75);")
+  expect_equal(x$GLOBAL[1], "/**")
+  expect_equal(x$GLOBAL[3], "*/")
+  expect_equal(x$PARAM, "THETA1 = 1")
+})
+
+
+code_convert_pow_1 <- '
```
kyleam: `code_convert_pow_1` and `code_convert_pow_2` (line 1006) are only used in one block. Why define them outside the block?

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  mod <- mcode('power', code_convert_pow_2, compile = FALSE)
+  x <- readLines(mod@shlib$source)
+  x <- x[grepl("d =", x, fixed = TRUE)]
+  expect_length(x, 2)
+  expect_match(x, "pow(a, b)", fixed = TRUE, all = TRUE)
+})
+
+rm(code_convert_pow_1, code_convert_pow_2)
+
+test_that("convert_pow handles comparison operators in exponent", {
```
kyleam, line 1023: [ just a question for my own understanding ]

Hmm, these conditions would either go to 0 or 1. Would you ever expect this in real code? (Same comment for "logical operators" case couple of blocks down.)

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  expect_equal(convert_pow("THETA(2)**(RACE!=3)"),   "pow(THETA(2), RACE!=3)")
+  expect_equal(convert_pow("THETA(2)**(RACE>=3)"),   "pow(THETA(2), RACE>=3)")
+  expect_equal(convert_pow("THETA(2)**(RACE<=3)"),   "pow(THETA(2), RACE<=3)")
+  expect_equal(convert_pow("THETA(2)**(RACE>3)"),    "pow(THETA(2), RACE>3)")
+  expect_equal(convert_pow("THETA(2)**(RACE<3)"),    "pow(THETA(2), RACE<3)")
+})
+
+test_that("convert_pow handles comparison operators in assignment prefix", {
+  expect_equal(
+    convert_pow("if(a==3) b = THETA(2)**(RACE)"),
```
kyleam: Why do all these cases put parentheses around `RACE`?

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+    mcode("int-div-env-2", code, compile = FALSE), 
+  )
+})
+
+# convert_fort_if -----------------------------------------------------
+
+fi <- convert_fort_if
+
+test_that("convert_fort_if: block form with Fortran operator", {
+  x <- c("IF(WT.GE.70) THEN", "  CL = THETA(1)", "ENDIF")
```
kyleam, line 1205: NM-TRAN accepts no space between the condition and `THEN` (produces identical FSUBS). That's parsed correctly but may be worth adding a test case.

tests/testthat/test-modspec.R
```diff
@@ -780,3 +780,679 @@ test_that("Invalid item in $SET generates error", {
...
+  expect_equal(fi("else"),              "} else {")
+})
+
+test_that("convert_fort_if: indentation preserved", {
+  expect_equal(fi("  IF(A.GT.0) THEN"), "  if(A>0) {")
+  expect_equal(fi("  ENDIF"),           "  }")
+})
+
+test_that("convert_fort_if: operators converted in body lines", {
+  expect_equal(fi("  CL = CL * (SEX.EQ.1)"), "  CL = CL * (SEX==1)")
```
kyleam: "body lines" - the `convert_fort_if` happens with only line scope/context, so it's really any code, isn't it? Not just the body within the if/else context.

src/dsl-preprocess.cpp
kylebaron: You are right; `.EQN.` and `.NEN.` in `$PK` fails to compile. I will remove them from the language.

src/dsl-preprocess.cpp
kylebaron: Confirming that NONMEM respects `&` to continue the line; will figure out what to do with it.

src/dsl-preprocess.cpp
kylebaron: I didn't appreciate this behavior but ok if it fails to convert and lets the user know; ok to just ask user to write pow() in this case.

inst/maintenance/unit-cpp/test-cpp.R
kylebaron: I believe it came from difficulty with CI and the potential that `LinkingTo` packages won't be around at run time (even if they were at install time). But you're right, in this case we do import from `Rcpp` (which is a hack in itself); so that should be enough to have the package there at run time.

kylebaron: I will remove and see if CI still works.

src/dsl-preprocess.cpp
kylebaron: Removed the numbering.

src/dsl-preprocess.cpp
kylebaron: Removed EQN and NEN

src/dsl-preprocess.cpp
kylebaron: Removed NOT, TRUE, and FALSE

tests/testthat/test-modspec.R
kylebaron: Test case added.

tests/testthat/test-modspec.R
kylebaron: Just a personal habit; I thought I had some (very minor) difficulties with the formatting while writing the code when it's inside the test (indenting etc). Moved the code inside the test and removed the `rm()` call.

tests/testthat/test-modspec.R
kylebaron: Keeping the test.

src/dsl-preprocess.cpp
kylebaron: I did fix semicolon handling on this:
```c
if(IFLAG > 0)
  a = b;
```

kylebaron: `convert_semicolons("a = b // comment")` is now handled.

src/dsl-preprocess.cpp
kylebaron: Done as suggested.

src/dsl-preprocess.cpp
kylebaron: Added this mapping.

src/dsl-preprocess.cpp
kylebaron: Right; while I hope `**` is as robust as possible, this and pretty much everything else in this PR assumes you're writing straightforward mrgsolve. You can do some crazy stuff, tapping all sorts of depths of C++, but I'd ask the user in that case to stick to standard C++. That said, the times when I've invoked some non-trivial C++, the pattern is to not do that in the model code block itself; we develop header files (1) for good organization and (2) to purposely fence that off from the more model-like blocks. I know we can't enforce this, but it's been an guiding principle for a long time and one of prices you pay to let any / all C++ in the door.

src/dsl-preprocess.cpp
kylebaron: Yeah; this started as _only_ converting pow; so it made sense at the time; I'll update the name.

src/dsl-preprocess.cpp
kylebaron: Handling negative numbers added.

src/dsl-preprocess.cpp
kylebaron: Changed to `**`: 

>The cleanest option would be a dedicated string that can never appear in user source code — something like "__pow__". It's unambiguous, self-documenting, and impossible to confuse with any C++ operator or user-written token since the grammar only produces it internally via the ** rule.
>
>The alternative would be keeping it as a string but using "**" itself — store what the user wrote, emit as pow(). That's arguably the most readable since the AST node directly reflects the source syntax.
>
>I'd lean toward "**" — it requires zero explanation when reading the BinaryExpr node, and there's no risk of collision since ** isn't a valid C++ operator.

This was the reflection on using `^`: 

>Yes, ^ is legitimate C++ — it's the bitwise XOR operator. So a ^ b in model code (unlikely but valid) would pass through convert_pow unchanged, which is correct behavior. But if someone wrote a ^ b intending XOR, there's no confusion because ^ is never parsed as an input token by the grammar — it only appears internally after ** is parsed.
>
>The real question is whether the AST-internal ^ could leak out incorrectly. Looking at the emitter, every ^ node is unconditionally emitted as pow(...), so there's no path where an internal ^ survives into the output as a literal ^. The only ^ that can appear in the output is one that was in the original source and was never parsed (i.e., the parser failed or skipped that expression), in which case the original line is returned unchanged.
>
>So the current design is safe. A more explicit internal token like "**" stored as-is, or a dedicated enum, would be cleaner — but since the AST is entirely internal and never serialized except through the emitter, the ^ sentinel works without risk of confusion.

src/dsl-preprocess.cpp
kylebaron: > Probably fine to represent, but NM-TRAN rejects this:
Going to retain `a ** -2`; it seems reasonable code. The `**` is going to be in effect regardless of whether or not you are coming from NONMEM.

kylebaron: Confirmed right-associativity; clarified comment in the code.

src/dsl-preprocess.cpp
kylebaron: Yeah; I had this implemented for a while as a bunch of preprocessor directives, subbing stuff out and it worked pretty well; I thought about just doing a simple implementation of this in R, scanning by line. So I guess the C++ implementation isn't too far from that.

src/dsl-preprocess.cpp
kylebaron: Yeah; I was sort of thinking about my style of preferring to keep the operator at the end of the line and I only handled the handful of the basic ops. 

I'm very tempted to try to handle this some way; but I think just letting the error happen and not assuming anything is the safest way to go. 

1. I almost never see people using `&`; doesn't mean a lot but practically should be rare to see it
2. I could see stripping a lone `&` off the end of the line; the presumption whenever we use the `semicolons` plugin is that we are coming from `NONMEM` so that should be the natural assumption - the `&` means continuation; if someone did sneak some C++ in, they should be using `&&`; if they use `&` and get tripped up, use `&&` instead to avoid stripping.
3. I had another option in mind, but lost the handle on that one while typing out 2. 

What do you think @kyleam ? Let it fail if you leave `&` at the end of the line?

tests/testthat/test-modspec.R
kylebaron: Not sure; removed parens from some of the tests; still passing.

tests/testthat/test-modspec.R
kylebaron: Not sure what happened there; I put "converted on rhs"; that was the goal of the test.

src/dsl-preprocess.cpp
kyleam: > Let it fail if you leave `&` at the end of the line?

That'd be my vote.

tests/testthat/test-modspec.R
kylebaron: But this is correct; NONMEM will choke on this construct.

kylebaron: Ok; I rolled this back; it should let this construct (`SEX.EQ.1`) pass through when on RHS.

kylebaron: Added test showing this passes through now.

tests/testthat/test-modspec.R
kylebaron: Confirmed that NONMEM will not allow this syntax; it was me thinking in R land. I will roll this back.

kylebaron: But let me see if C++ will allow it.

kylebaron: I'm going to keep it since C++ handles it the right way. This is definitely could be in real code.

kylebaron: @kyleam If I'm reading the comment right, the logical operators below are the NONMEM/Fortran constructs; i that case, I think it makes sense to nuke that since that is only invoked with `nm-vars` and we are insisting people stay in the vicinity of NONMEM in that case.

tests/testthat/test-modspec.R
kylebaron: Could you help me with what went wrong; the code in the test renders to this and the model compiles:

```c
// GLOBAL START USER CODE:
  /** 
  This model is incredible.
  */
  double foo = 123;


// MAIN CODE BLOCK:
__BEGIN_main__
MRGSOLVE_WARN_UNUSED_VAR_NO
const double& THETA1 = _THETA_[0];
const double& WT = _THETA_[1];
MRGSOLVE_WARN_UNUSED_VAR_YES
////////////////////////////////////////////////////////////////////////////////
  CL = THETA1*pow(WT/123, 0.75);
////////////////////////////////////////////////////////////////////////////////
__END_main__
```

src/dsl-preprocess.cpp
kylebaron: Will do; thanks @kyleam .

tests/testthat/test-modspec.R
kyleam: > If I'm reading the comment right, the logical operators below

I was referring to the tests on 1050:

https://github.com/metrumresearchgroup/mrgsolve/blob/03460cb6b329f0ffa1765b1252af9f18c346b2df/tests/testthat/test-modspec.R#L1050-L1059

Those use `THETA(n)`, which IIUC are NONMEM-inspired macros but always in effect (not tied to nm-vars).

src/dsl-preprocess.cpp
kylebaron: That was hold-over code from when it was an error (I think); removed.

src/dsl-preprocess.cpp
kylebaron: Thinking ....

kylebaron:
@kyleam Here's the change set so far: https://github.com/metrumresearchgroup/mrgsolve/pull/1362 

This helps me keep track of the changes; I don't know if it helps you at all (sort of guessing you don't really need it). Just let me know. I can merge once we're ready to hand back to you and go from there.

src/dsl-preprocess.cpp
kylebaron: @kyleam You're suggesting we just skip the `if (needs_parens_...)` part and always put parens in there?

inst/maintenance/unit-cpp/test-cpp.R
kylebaron: It works.

tests/testthat/test-modspec.R
kyleam: The test case says "handles c-style comment" and then shows a multi-line comment.  I'm saying I find that description misleading, because as discussed elsewhere, everything here is lined-based and doesn't have wider context (so it botches mutli-line comments and treats the inner text as regular code).  

So it says it handles c-style comments, but if you put a "3 ** 4" into the comment of the test, it gives the warning I showed (tested on 03460cb6):

```
/**
This model is incredible. 3 ** 4
*/
```

tests/testthat/test-modspec.R
kylebaron: IC; right, those would stay as viable code in mrgsolve. Please let me know if I'm not following something in the comment.

tests/testthat/test-modspec.R
kylebaron: OIC; I got it now. Thanks for clarifying. I'll fix the description there.

kylebaron: 1. Clarified test description around c-style comment
2. Added test demonstrating that text in the middle of c-style comment is subject to conversion and possible failure

kylebaron: Ok; this is staying in its lane now; opened up and issue to consider stripping those comment blocks out at the get-go, prior to processing any of the code. Will do that under separate cover once this PR is in good shape.  See #1363

src/dsl-preprocess.cpp
kyleam: Hmm, it wasn't that concrete of a suggestion, just the thought that you could probably get by with less logic (and hopefully fewer gotchas) by explicitly wrapping more things.  I don't think you'd want to drop the `needs_parens_*` conditions entirely, because then you'd start wrapping atoms in parentheses.  Playing around with it a bit, I'd say something like the diff below.

But despite being "tempted" (and thinking it's worth mentioning, because it gets rid of a good chunk of logic), I'm not really sold on the idea.  And I don't think the tests flagged any issues in the area.  So I'd leave it be for now.

<details>
<summary>diff</summary>

```diff
diff --git a/src/dsl-preprocess.cpp b/src/dsl-preprocess.cpp
index c34c6024..14d67016 100644
--- a/src/dsl-preprocess.cpp
+++ b/src/dsl-preprocess.cpp
@@ -92,33 +92,13 @@ struct FuncCall {
 // 2.  Emitter  (AST -> C++ string)
 // ---------------------------------------------------------------------------

-static int precedence(const std::string& op) {
-  if (op == "||") return 1;
-  if (op == "&&") return 2;
-  if (op == "==" || op == "!=" ||
-      op == "<"  || op == "<=" ||
-      op == ">"  || op == ">=") return 3;
-  if (op == "+" || op == "-") return 4;
-  if (op == "*" || op == "/") return 5;
-  return 0;  // "^" / unknown — not used for parens decisions (pow is a function)
-}
-
 static const ast::BinaryExpr* as_binop(const ast::Expr& e) {
   return boost::get<ast::BinaryExpr>(&e);
 }

-static bool needs_parens_left(const ast::Expr& child, const std::string& parent_op) {
-  const ast::BinaryExpr* b = as_binop(child);
-  if (!b || b->op == "^") return false;
-  return precedence(b->op) < precedence(parent_op);
-}
-
-static bool needs_parens_right(const ast::Expr& child, const std::string& parent_op) {
+static bool needs_parens(const ast::Expr& child, const std::string& parent_op) {
   const ast::BinaryExpr* b = as_binop(child);
-  if (!b || b->op == "^") return false;
-  int cp = precedence(b->op);
-  int pp = precedence(parent_op);
-  return cp < pp || (cp == pp && (parent_op == "-" || parent_op == "/"));
+  return b && b->op != "^";
 }

 struct Emitter : boost::static_visitor<std::string> {
@@ -145,8 +125,8 @@ struct Emitter : boost::static_visitor<std::string> {
     if (b.op == "^") {
       return "pow(" + l + ", " + r + ")";
     }
-    if (needs_parens_left(b.lhs, b.op)) l = "(" + l + ")";
-    if (needs_parens_right(b.rhs, b.op)) r = "(" + r + ")";
+    if (needs_parens(b.lhs, b.op)) l = "(" + l + ")";
+    if (needs_parens(b.rhs, b.op)) r = "(" + r + ")";
     return l + b.op + r;
   }
```

</details>

kyleam:
> I can merge once we're ready to hand back to you and go from there.

That works for me, thanks.

src/dsl-preprocess.cpp
kylebaron: I tweaked this to crop mention of C-style; I think it's targeting `//` but please let me know if I'm not following.

R/RcppExports.R
```diff
@@ -5,6 +5,67 @@ DEVTRAN <- function(parin, funs, data, idata, mod) {
...
+}
+
+#' Convert Fortran-style IF/THEN/ELSE/ENDIF to C++
+#'
+#' Translates Fortran block-form and single-line IF constructs to C++ in each
+#' element of \code{code}.  Fortran relational and logical operators
+#' (\code{.GE.}, \code{.LE.}, \code{.GT.}, \code{.LT.}, \code{.EQ.},
+#' \code{.NE.}, \code{.AND.}, \code{.OR.}) are converted everywhere they
+#' appear. Matching is
+#' case-insensitive.
```
kyleam: doc nit: awkward line break

src/dsl-preprocess.cpp
```diff
@@ -0,0 +1,903 @@
...
+  int depth = 0;
+  for (size_t i = 0; i < s.size(); ++i) {
+    if      (s[i] == '(') ++depth;
+    else if (s[i] == ')') --depth;
+    else if (s[i] == ch && depth == 0) return i;
+  }
+  return std::string::npos;
+}
+
+static size_t find_semi_depth0(const std::string& s) {
```
kyleam: Fine as is, but given it's called once, I don't see much gain in having `find_semi_depth0` rather than inlining `find_at_depth0(s, ';')`.

kyleam (approved):
The updates, in particular the new tests, look as I expected. In general, I think this is in a good state to merge in. This area is full of gotchas but

-   as discussed, in the context of mrgsolve cpp code, what's implemented will be sufficient for most cases

-   when bad cases come up, all the knobs are in place for the user to disable them, and the code in dsl-preprocess.cpp should provide a good starting point for account for the case, if it seems worth doing for the particular case

src/dsl-preprocess.cpp
kyleam: My comment was pointing out a confusion that I had while reviewing the code. The comment says "C-style", but there are of course multiple variants and the code does not implement support for the multi-line variant. I don't think it's unreasonable for a reader to expect the code to provide full support for variants, rather than it doing a context-limited parsing of a line, because the code is pulling a full-featured library to do parsing.

Line-based parsing is probably sufficient in the context of mrgsolve (as discussed in [other](https://github.com/metrumresearchgroup/mrgsolve/pull/1360#discussion_r3059650330) [comments](https://github.com/metrumresearchgroup/mrgsolve/pull/1360#discussion_r3060248861)), but it doesn't mean it won't trip up readers trying to understand the code. (Of course it could have been just *this* reader.)

In any case, I'm not asking for anything in particular here, though I certainly don't think it would hurt for dsl-preprocess.cpp to have some high-level comments from you explaining what the approach is, along the lines of what you said [here](https://github.com/metrumresearchgroup/mrgsolve/pull/1360#discussion_r3060248861) and [here](https://github.com/metrumresearchgroup/mrgsolve/pull/1360#discussion_r3060456488).

R/RcppExports.R
kylebaron: Re-flowed the comment.

src/dsl-preprocess.cpp
kylebaron: Ah; I see now. That makes sense. It is now inlined.

src/dsl-preprocess.cpp
kylebaron: Leaving for now, noting review discussion.

src/dsl-preprocess.cpp
kylebaron: Maybe calling it a dsl preprocessor was going too far. 

For current and future context: 

I'm already doing a bunch of interrogation and processing of the C++ code using R with regular expressions. All of that is line-based for simplicity of not having to work around newline and the fact is that the whole model file is split by newline right out of the gate when the list of code lines comes out of `modelparse()`. As I noted in the main PR comment, the code in `dsl-preprocess.cpp` was a more powerful / reliable version of what I might have done processing line by line with the R code. I'm very happy to be able to make this available with line-by-line processing limitation. mrgsolve will always be in a precarious situation syntax-wise because one of the desire to allow unlimited (or virtually unlimited) access to straightforward C++ while also wanting to take a big edge off for users who just want to write some equations. We've edged into building bridges to NONMEM syntax as well which complicates things too. I don't know how I would have done evtools without having the freedom to let it be implemented in plain old C++.  Finally, we need to remember that this isn't a regular C++ program; it needs to be a set of functions in a single translation unit that mrgsolve has to be able to call (from R and then C++) at the right time with no knowledge about the model specifics at run time. Along these lines, there is a major problem with the error messages that get emitted when there is an error in the C++ code; the compiler sends the error to stderr, but it can be littered with stuff injected by mrgsolve, line numbers don't match up, etc. That's always been there and will always be there in some form or another b/c we're trying to walk this line. So (1) it's weird from the very start and (2) different users / different contexts will have different needs and desires for operating in that weirdness. IMO, #1332 was a huge improvement in finding a way to let users reference "parameters" and "compartments" in their model by the names they decide; those global preprocessor directives worked great at first, but the more we got into writing complex C++ code the more we had to dodge collisions. So that was huge. And it also illustrates the line we are trying to walk, acknowledging everyone is working from R, most want simplicity, a lot are coming from NONMEM, and some want to be able to go all out C++-wise.

kylebaron:
Thanks for the detailed review, @kyleam. I know this was a big ask and you really dug into how it was all working and the review really strengthened this addition to the code base.

PR #1361, opened 2026-04-04
Consistent pk parameters

kylebaron:
This looks really bad; but hardly any tests and no examples needed changing. Nevertheless, will be a breaking change. I'm thinking next release is v2.0.

The `pk1`, `pk2` and `pk3` models are all in closed form; they have structural parameters that match what NONMEM expects. We switch to those parameter names for (almost) the whole `pk*` series and `irm*`, `tmdd` and 
`emax`. Going to leave other models alone since they have different parameterization or the context is just different.  

In addition to clearances and volumes, we also change `KA1` to `KA` and `EV1` to `EV` when there are two depot compartments.

This also puts the `@input` tag on the key parameter names; that would allow the user to inventory the simulation data set and check for consistency of parameters. 

@kyleam This is the second to last "big" thing I had planned in the current development cycle.


R/modlib.R
```diff
@@ -96,32 +96,32 @@ modlib_list <- function() {
...
-#' - `EV1`, `EV2`: extravascular dosing compartments
+#' - `EV`, `EV2`: extravascular dosing compartments
 #' - `CENT`: central PK compartment
 #' - `PERIPH`: peripheral PK compartment
 #' - `PERIPH2`: peripheral PK compartment 2
 #' - `RESP`: response PD compartment (irm models)
 #'
 #' @section Output variables:
-#' - `CP`: concentration in the central compartment (`CENT/VC`)
+#' - `CP`: concentration in the central compartment (`CENT/V2`)
```
kyleam: For pk1 and pk1cmt, is it `CENT/V` rather than `CENT/V2`?

https://github.com/metrumresearchgroup/mrgsolve/blob/e9d5cb1b734b1622fcd6397736d32d1f87b20114/inst/models/pk1.cpp#L11

https://github.com/metrumresearchgroup/mrgsolve/blob/e9d5cb1b734b1622fcd6397736d32d1f87b20114/inst/models/pk1cmt.cpp#L35

R/modlib.R
```diff
@@ -167,11 +167,11 @@ object_dir <- function() {
 #' See [modlib_details] for more detailed descriptions of parameters and
 #' compartments.
 #'
-#' The `pk1cmt` model is parameterized in terms of `CL`, `VC`, `KA1` and
-#' `KA2` and uses compartments `EV1`, `EV2`, and `CENT`. The `pk2cmt` model
-#' adds a `PERIPH` compartment and parameters `Q` and `VP` to that of the
+#' The `pk1cmt` model is parameterized in terms of `CL`, `V2`, `KA` and
```
kyleam: Should this say `V` rather than `V2`?

https://github.com/metrumresearchgroup/mrgsolve/blob/e9d5cb1b734b1622fcd6397736d32d1f87b20114/inst/models/pk1cmt.cpp#L21

R/modlib.R
```diff
@@ -96,32 +96,32 @@ modlib_list <- function() {
...
 #'   second extravascular compartment (1/time)
 #' - `CL`: clearance (volume/time)
-#' - `VC`: volume of distribution, central compartment (volume)
-#' - `VP`: volume of distribution, peripheral compartment (volume)
-#' - `VP2`: volume of distribution, peripheral compartment 2 (volume)
+#' - `V2`: volume of distribution, central compartment (volume)
+#' - `V3`: volume of distribution, peripheral compartment (volume)
+#' - `V4`: volume of distribution, peripheral compartment 2 (volume)
 #' - `Q`: intercompartmental clearance (volume/time)
-#' - `Q2`: intercompartmental clearance 2 (volume/time)
```
kyleam, line 117: Should this include `Q3`?

https://github.com/metrumresearchgroup/mrgsolve/blob/e9d5cb1b734b1622fcd6397736d32d1f87b20114/inst/models/pk3.cpp#L4

https://github.com/metrumresearchgroup/mrgsolve/blob/e9d5cb1b734b1622fcd6397736d32d1f87b20114/inst/models/pk3cmt.cpp#L13

kyleam (commented):
Noting what I think are a few doc/code mismatches, but in each case I could just be getting turned around.

R/modlib.R
kylebaron: Ugh; I was afraid of this. I can take another pass through everything. But I'd prefer to just nuke this documentation; I gave up on trying to make all the entries for new models b/c they should be self-explanatory. Will review again.

R/modlib.R
kylebaron: This is complicated because the help file doesn't address just one model; so that expression is different for a bunch of different models. 

I'm going to remove the derivations and just list the different clearances and volumes. There will be some additional parameters coming through (missing from the previous docs).

R/modlib.R
kylebaron: Yes; and `V`: 


```r
 #' @section Output variables:
-#' - `CP`: concentration in the central compartment (`CENT/V2`)
+#' - `CP`: concentration in the central compartment
 #' - `RESP`: response (emax model)
 #'
 #' @section PK parameters:
 #' - `KA`, `KA2`: first order absorption rate constants from first and
 #'   second extravascular compartment (1/time)
 #' - `CL`: clearance (volume/time)
+#' - `V`: volume of distribution (volume)
 #' - `V2`: volume of distribution, central compartment (volume)
 #' - `V3`: volume of distribution, peripheral compartment (volume)
 #' - `V4`: volume of distribution, peripheral compartment 2 (volume)
 #' - `Q`: intercompartmental clearance (volume/time)
+#' - `Q3`: intercompartmental clearance (volume/time)
 #' - `Q4`: intercompartmental clearance 2 (volume/time)
 #' - `VMAX`: maximum rate, nonlinear process (mass/time)
 #' - `KM`: Michaelis constant (mass/volume)
-#' - `K10`: elimination rate constant (1/time); `CL/V2`
+#' - `K10`: elimination rate constant (1/time)
 #' - `K12`: rate constant for transfer to peripheral compartment from
-#'   central (1/time); `Q/V2`
+#'   central (1/time)
 #' - `K21`: rate constant for transfer to central compartment from
-#'   peripheral (1/time); `Q/V3`
+#'   peripheral (1/time)

```

R/modlib.R
kylebaron: Yes; the doc was wrong in the previous iteration.

kyleam (approved):

kylebaron:
@kyleam Thanks for reviewing that. I have now doubt it was very annoying. I did fix the stuff you picked up and found some other nits along the way. I think as far as it goes, the this recovers from the new problems that cropped up with my refactoring and fixes some existing issues. 

I'm happy to keep going, merging this in. Or just cut the cord on this breaking release. Any thoughts one way or another?

kylebaron:
Didn't see you already approved. Will keep this around for the time being. Maybe revisit when `v3.0.0` comes out :)

PR #1364, opened 2026-04-10
Add to dsl-preprocess to strip block comments

kylebaron:


kylebaron:
Not going to pursue this now.

PR #1366, opened 2026-04-13
Rstudio addin to preprocess code 

kylebaron:
This extends #1360, providing an Rstudio plugin to enable selective processing of code when the model file is open in the IDE. A lot of people use Rstudio and this would let them convert code (1) preserving the changes in the file and  (2) let them see what is happening rather than always under the hood. 

I eventually want to take this idea into some pseudo translation routine, independent of Rstudio. But for now, I think getting people to use the addin will help us see how this is all working, what sort of tweaks might be needed and how we can most effectively give people the mrsolve starter material from a nonmem control stream.

Code must be selected to be processed; error otherwise. 

If the selected code contains block ID markers, the code is split in to a spec list using newly-developed `modelsplit()`, which is a specific call to `modelparse()` set up to preserve code. After processing, the list of code is  put back into a single character vector using `modelunsplit()`.  If no block ID markers are found, the code lines are processed. Some refactoring of `modelparse()` was also performed to support `modelsplit()` and `modelunsplit()` as well as improving the flow and documentation. 

1. Convert NM `runs convert_fort_if()` and `convert_semicolons()`
2. Add semicolons runs `convert_semicolons()`; unlike the plugin, this does not require `nm-vars`
3. Convert pow runs `convert_pow()`

`convert_pow()` is known to eat spaces and it also inserts space to the right of `=`. This isn't a problem under the hood, but isn't great when the function is modifying users code in the file. Therefore, I added an additional call to ensure balanced spaces around `=`.  This seems ok to me and the limitation will be documented. My hope is that people just leave `**` in their code and rely on conversion under the hood. 

Because this is working on code as it stands in the file (with no scraping of comments of @ directives), I needed to update some of the `dsl-preprocess.cpp` language to accommodate this. 

While testing, I also realized that we didn't handle leading `;` correctly; it should be converted to `//` ; this is analogous to how we handle existing `;` that happens at the end of a line (leave `;` and add `//` after).  This is now handled. 

While I want to keep the `semicolons` addin, my hope is that people will just call the plugin, put the semicolons into their code, hit save and be done.


R/dsl-preprocess-addin.R
```diff
@@ -0,0 +1,71 @@
...
+  text <- sel$text
+  if(!nchar(text)) { 
+    # error and warning don't work quite the way I want
+    msg <- c(x = "No text selected; please select the code to process.")
+    inform(msg); 
+    return(invisible(NULL))
+  }
+  lines <- strsplit(text, "\n", fixed = TRUE)[[1]]
+  result <- apply_fn(text, lines)
+  # Don't each blank line at end of selection
```
kyleam: `s/each/eat/` ?

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
-modelparse <- function(txt, split=FALSE, drop_blank = TRUE, 
-                       comment_re=c("//", "##")) {
-  
-  ## Take in model text and parse it out
+#' Parse model specification text
+#' 
+#' @param txt model specification text.
+#' @param split logical; if `TRUE`, `txt` will be split on `\n` before 
+#' processing.
+#' @param drop_blank logical; `TRUE` if blank lines wilil be dropped.
```
kyleam: `s/wilil/will/`

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
+#' @return `modelsplit()` returns a named list of character vectors, one element
+#' per block. `modelunsplit()` returns a character vector of model code.
+#'
+#' @seealso [modelparse()] for a more destructive parse that drops blank lines
+#' and comments.
+#'
+#' @examples
+#' file <- file.path(modlib(), "pk1.cpp")
+#' 
+#' x <- modelsplit(readLines(file))
```
kyleam: Should this use `split = FALSE` given `readLines` already splits on newlines?

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
+#' 
+#' modelunsplit(x)
+#' 
+#' @md
+#' @export
+modelsplit <- function(x, split = TRUE) {
```
kyleam: I realize this is just relying the `split` arg to `modelparse`, but I think "split" is a confusing parameter name for function named `modelsplit`.

All the current callers pass `split = FALSE`, so it's not obvious to me that it's worth adding the parameter at this point.

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
+#' @param x for `modelsplit()`, model specification text (character vector or
+#' single string); for `modelunsplit()`, a list returned by `modelsplit()`.
+#' @param split logical; if `TRUE`, a single string is split on newlines before
+#' parsing.
+#' 
+#' @return `modelsplit()` returns a named list of character vectors, one element
+#' per block. `modelunsplit()` returns a character vector of model code.
+#'
+#' @seealso [modelparse()] for a more destructive parse that drops blank lines
+#' and comments.
```
kyleam: I find this wording confusing given `modelsplit` calls `modelparse` underneath (so whether `modelparse` is destructive depends on which arguments are passed).

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
+#' @rdname modelsplit
+#' @export
+#' @keywords internal
+modelunsplit <- function(x) {
+  bloc <- attr(x, "blockmatch")
+  if(is.null(bloc)) stop("cannot unsplit model specification list.")
+  header <- attr(x, "header")
+  for(i in seq_along(bloc)) {
+    sep <- ifelse(x[[i]][[1]] == "", "", " ")
+    x[[i]][[1]] <- paste0(bloc[[i]], sep, x[[i]][[1]])
```
kyleam, line 509: I think `x[[i]][[1]]` is always a scalar. In that case, it'd be clearer to avoid the vectorized `ifelse` (with, e.g., `if (x[[i]][[1]] == "") "" else " "`).

R/modspec.R
```diff
@@ -352,73 +356,211 @@ convert_semicolons <- function(x) {
...
+  for(i in to_convert) {
+    x[[i]] <- convert_pow(x[[i]], names(x)[i])
+  }
+  x
+}
+
+# Pad only the first (assignment) = on each line; compound operators like
+# !=, <=, >= are left alone.
+add_leading_eq_space <- function(x) {
+  sub("(?<![!<>=])\\s*=\\s*(?!=)", " = ", x, perl = TRUE)
```
kyleam: This would mishandle a `/=` condition.

For example

```
IF(X/=3) Y=1.3
```

becomes

```
IF(X/ = 3) Y=1.3
```

> \`convert\_pow()\` is known to eat spaces and it also inserts space to the right of \`=\`. This isn't a problem under the hood, but isn't great when the function is modifying users code in the file. Therefore, I added an additional call to ensure balanced spaces around \`=\`. This seems ok to me and the limitation will be documented. My hope is that people just leave \`\*\*\` in their code and rely on conversion under the hood.

Making this adjustment at this level is fine with me, though it does seem a bit odd to handle this here rather changing `convert_pow` to behave in a way that works as you want for this case too.

R/dsl-preprocess-addin.R
```diff
@@ -0,0 +1,71 @@
...
+  sel <- ctx$selection[[1]]
+  text <- sel$text
+  if(!nchar(text)) { 
+    # error and warning don't work quite the way I want
+    msg <- c(x = "No text selected; please select the code to process.")
+    inform(msg); 
+    return(invisible(NULL))
+  }
+  lines <- strsplit(text, "\n", fixed = TRUE)[[1]]
+  result <- apply_fn(text, lines)
```
kyleam: I didn't spot any `addin_apply_*` function that used the `text` argument. Assuming I'm not overlooking a spot, perhaps drop the arg until it's needed?

NAMESPACE
```diff
@@ -150,6 +150,8 @@ export(mcode_cache)
 export(modMATRIX)
 export(modelparse)
 export(modelparse_rmd)
+export(modelsplit)
+export(modelunsplit)
```
kyleam: Do `modelsplit` and `modelunsplit` need to be exported?

R/modspec.R
kylebaron: Removed; convert_pow leaves all other lines alone; this was a way to try to even the output. I tried several different thing but going to remove.

inst/rstudio/addins.dcf
```diff
@@ -0,0 +1,16 @@
+Name: Convert NM
+Description: Converts Fortran power, IF constructs, and inserts semicolons into mrgsolve model DSL code, then adds spaces around the assignment operator. Requires a selection.
```
kyleam: > Converts Fortran power

`addin_convert_nm` doesn't do the pow conversion, right?

https://github.com/metrumresearchgroup/mrgsolve/blob/cdc281cc0ea439edbdd332492ec894a046a2a1c7/R/dsl-preprocess-addin.R#L21-L27

kylebaron: Right; I pared that back because of the formatting issues. I sort of see the `**` operator becoming  part of standard mrgsolve.

R/dsl-preprocess-addin.R
kylebaron: Dropped that arg.

R/modspec.R
kylebaron: Yeah;  I think the only context were it would be used is when writing code inline as an R string; the split part was sort of a convenience there.

R/mread.R
```diff
@@ -457,8 +457,13 @@ mread <- function(model, project = getOption("mrgsolve.project", getwd()),
...
-  rd <- arrange_rdefs(rd) 
-  
+  rd <- arrange_rdefs(rd)
+
+  # Convert pow just before writing the file
+  if(mread.env$convert_pow) {
+    spec <- convert_pow_spec(spec, mread.env$incoming_names)
+  }
+
   ## Collect all code to be written to the different blocks
```
kylebaron, line 467: Pass in "original" block names for more accurate warning.

R/dsl-preprocess-addin.R
```diff
@@ -0,0 +1,79 @@
...
+    result <- modelsplit(lines)
+    withCallingHandlers({
+      result <- convert_pow_spec(result)
+    }, warning = addin_warning)
+    result <- modelunsplit(result)
+  } else {
+    withCallingHandlers({
+      result <- convert_pow(lines)
+    }, warning = addin_warning)
+  }
```
kylebaron, line 72: `convert_pow()` might warn if it can't parse a line.

R/modspec.R
```diff
@@ -352,73 +356,187 @@ convert_semicolons <- function(x) {
...
+
+# Apply ** to pow conversion to the right blocks; used in addin
+convert_pow_spec <- function(x, block_names = names(x)) {
+  to_convert <- which(names(x) %in% GLOBALS$PRE_PROC_BLOCKS)
+  to_convert_names <- block_names[to_convert]
+  for(i in to_convert) {
+    x[[i]] <- convert_pow(x[[i]], to_convert_names[i])
+  }
+  x
 }
```
kylebaron, line 540: Allow to pass in the "original" block names; needed b/c now we're using `convert_pow_spec()` under the hood to do this when running `mread()`.

src/dsl-preprocess.cpp
```diff
@@ -487,6 +487,13 @@ static std::string convert_pow_line(const std::string& line) {
...
   }
 
+  // A space will be added to the right of '='; ensure one also appears to the
+  // left so we don't produce asymmetric output like "CL= pow(...)".
+  if (prefix.size() >= 2 && prefix.back() == '=' &&
+      prefix[prefix.size() - 2] != ' ') {
+    prefix.insert(prefix.size() - 1, 1, ' ');
+  }
+
   std::string sep = prefix.empty() ? "" : " ";
```
kylebaron, line 497: Minor clean up when processing lines; intended for when this and converted code goes back into the file (for now, the addin).

R/modspec.R
kylebaron: For future consideration / simplification: https://github.com/metrumresearchgroup/mrgsolve/issues/1379

kyleam (approved):

PR #1372, opened 2026-04-15
fix select warnings

kylebaron:
Struggling to find out where tidyselect is used on this one. 

```r
══ Warnings ════════════════════════════════════════════════════════════════════════════════════════════════════════════
1. inventory conditions (test-inventory-too.R:39:3) - Using an external vector in selections was deprecated in tidyselect 1.1.0.
ℹ Please use `all_of()` or `any_of()` instead.
  # Was:
  data %>% select(need)

  # Now:
  data %>% select(all_of(need))

See <https://tidyselect.r-lib.org/reference/faq-external-vector.html>.

══ DONE ════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 Done!
```


kyleam:
> Struggling to find out where tidyselect is used on this one.

Looking into it, think I've identified the spots that need updating.  I'll clean things up a bit and push here.

kyleam:
> I'll clean things up a bit and push here.

Done with 0fa3311a.  Please take a look.

R/mrgsims.R
```diff
@@ -292,7 +297,7 @@ select_sims <- function(.data, ...) {
     wstop("no output variables (compartments or captures) were selected.")  
   }
   vars <- unique(c(retain,vars))
-  .data@data <- dplyr::select(.data@data,vars)
+  .data@data <- .data@data[, vars, drop = FALSE]
```
kyleam, line 300: Works for me.

Sticking with `dplyr::select`, it could be `.data@data <- dplyr::select(.data@data, all_of(vars))`.

kyleam (approved):
Should be good to merge, assuming 0fa3311a looks good to you (and the CI passes after my push).

kylebaron:
Looks good; thank you for checking it out. This is all terrible stuff that should just go at some point. At least it's not giving us warnings.

PR #1378, opened 2026-04-21
Consolidate pow conversion to after all the other conversions

kylebaron:
@kyleam What do you think about this refactor?

In testing the addin stuff, I realized that `convert_pow()` gets called early on; so it could have to deal with more complicated code (like NONMEM code that will eventually get converted). 

This PR, I pull `convert_pow()` from the handler functions and do this conversion last, making sure it has the simplest, most uniform code to deal with. 

In terms of the plugin, that would have to be a potential limitation - that it's possible that `convert_pow()` could choke on some NONMEM constructs if Convert NM isn't called first. 

NB `convert_pow()` currently won't parse your example `IF(X/=3) Y=1.3**5`; I played with an update to `convert_pow()` to handle this but I started thinking it's just doing more than we need. With this refactor, it would let `convert_pow()` work when called under the hood because we're controlling the order, calling it last. If it crumps with the addin because the order wasn't right, I'd just say run NM addin first. And the long term goal with this is to write a (partial) converter, that would look to (1) generate a starter model file from an existing nonmem control stream with option to (2) take an existing mrgsolve model and update with the relevant model blocks from nonmem (e.g., $PK, $ODE, etc ... the stuff that actually changes as model development proceeds); so long term, we're going to control the conversion order.

### DIFF
--- a/R/dsl-preprocess-addin.R
+++ b/R/dsl-preprocess-addin.R
@@ -1,3 +1,9 @@
+addin_warning <- function(w) {
+  msg <- c(x = w$message)
+  inform(msg)
+  invokeRestart("muffleWarning")
+}
+
 addin_run <- function(ctx, apply_fn) {
   sel <- ctx$selection[[1]]
   text <- sel$text
@@ -55,10 +61,14 @@ addin_add_semicolons <- function() {
 addin_apply_convert_pow <- function(lines) {
   if(has_block_markers(lines)) {
     result <- modelsplit(lines)
-    result <- convert_pow_spec(result)
+    withCallingHandlers({
+      result <- convert_pow_spec(result)
+    }, warning = addin_warning)
     result <- modelunsplit(result)
   } else {
-    result <- convert_pow(lines)
+    withCallingHandlers({
+      result <- convert_pow(lines)
+    }, warning = addin_warning)
   }
   paste(result, collapse = "\n")
 }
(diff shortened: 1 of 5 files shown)


kyleam:
Makes sense to me, changes look fine.

inst/maintenance/unit/test-addin.R
```diff
@@ -74,10 +74,3 @@ test_that("Addin: convert nonmem converts IF/ELSE", {
   expect_match(out, "Z = 5;", fixed = TRUE)
   expect_match(out, "else if(WT == 80) {", fixed = TRUE)
 })
-
-test_that("convert_pow handles Fortran /= (not-equal) in condition", {
-  # /= is the Fortran not-equal operator; find_assign_eq must not treat it as
-  # an assignment so the RHS pow conversion can proceed
-  out <- mrgsolve:::convert_pow("IF(X/=3) Y=1.3**5")
-  expect_equal(out, "IF(X/=3) Y = pow(1.3, 5)")
-})
```
kylebaron, line 83: I had this test in and it was working, but reverted.

R/dsl-preprocess-addin.R
```diff
@@ -55,10 +61,14 @@ addin_add_semicolons <- function() {
...
+    withCallingHandlers({
+      result <- convert_pow_spec(result)
+    }, warning = addin_warning)
     result <- modelunsplit(result)
   } else {
-    result <- convert_pow(lines)
+    withCallingHandlers({
+      result <- convert_pow(lines)
+    }, warning = addin_warning)
   }
```
kylebaron, line 72: More complicated than I want; but need to do this so the warning surfaces on the console.

PR #1380, opened 2026-04-21
Release/2.0.0

kylebaron:
# mrgsolve 2.0.0

## Breaking Changes

- `data_set()` and `idata_set()` no longer accept `.subset`, `.select`,
  `object`, or `need` arguments. These arguments allowed filtering and
  column selection inside the call; that processing should now be done on the
  data frame before passing it to `data_set()` or `idata_set()`. (#1374)

- The `n` argument to `simeta()` and `simeps()` has been discouraged for a
  while and is now removed. Calling `simeta(n)` or `simeps(n)` to resimulate 
  a single ETA or EPS value is no longer supported. Use `simeta()` or `simeps()` 
  (no argument) to resimulate all values. (#1373)

- `knobs()`, `wf_sweep()`, and `render()` have been removed. These
  functions were previously deprecated. (#1369)

- The default value of `root` in `$NMXML` and `$NMEXT` has changed from
  `"working"` to `"cppfile"`. The root directory for locating NONMEM
  output files now defaults to the directory containing the model `.cpp` file
  rather than the R working directory. Pass `root = "working"` to restore the
  previous behavior; but users are _highly_ encouraged to use the new default.
  (#1368)

- `summarise.each()` is removed; it has not been practically reachable by the 
  user for several years (#1352).

- Several `modlib` model library models have had parameter and compartment
  names standardized (#1361):
  - The first extravascular compartment is now named `EV` (was `EV1`) in
    `pk1cmt`, `pk2cmt`, `irm1`-`irm4`, and `emax`.
  - The first absorption rate constant is now named `KA` (was `KA1`) in
    models that previously used numbered names.
  - Code that references these compartments or parameters by name (e.g.,
    `init(mod, EV1 = 0)` or `param(mod, KA1 = 1)`) will need to be updated.
  - We continue to discourage users from using these models in production 
    code. 

## New Features

- Models can now use `a ** b` syntax to represent `pow(a, b)`; this syntax is
  always available, without need to invoke a plugin (#1360).

- With the `nm-vars` plugin, mrgsolve now recognizes NONMEM / Fortran 
  `IF / ELSE / THEN` statements; this code can be left in your model and 
  mrgsolve will convert to appropriate C++ under the hood (#1360).

- `ERR(n)` is now recognized as an alias for `EPS(n)` when using the `nm-vars`
  plugin (#1367).

- New DSL preprocessing functions convert NONMEM/Fortran model code to 
  mrgsolve / C++ syntax (#1360, #1362, #1366):
  - `convert_pow()` converts Fortran `**` exponentiation to `pow()`.
  - `convert_fort_if()` converts Fortran `IF/THEN/ELSE/END IF` blocks to C++
    `if/else` syntax.
  - `convert_semicolons()` inserts missing trailing semicolons on assignment
    statements.
  - RStudio addins are provided to allow conversion of certain NONMEM source 
    code blocks to mrgsolve format in the RStudio editor window.

- Closed-form three-compartment linear models with or without depot compartments
  can now be implemented via `$PKMODEL` (#1345).

- `$PKMODEL` gains an `advan` input for selecting 1, 2, or 3 compartment
  models with analytical solutions; for example, setting `advan` to 2, 4,
  or 12 will give you 1, 2, or 3 compartment model in closed form with a
  depot compartment. When `advan` is specified, compartments with standard 
  names are automatically registered if neither `$INIT` nor `$CMT` are 
  found in the model file and no compartments are registered at the time 
  that `$PKMODEL` is processed (#1345).

- Two new `modlib` models, `pk3` and `pk3iv`, provide the corresponding
  pre-coded model files (#1345).

- `mrgx::assign()` is now available in the `mrgx` plugin; this is a convenience 
  function to help you get R objects from your model code back to R when the 
  simulation finishes (#1353).

- The `plot()` method for `mrgsims` objects gains a `fixy` argument to fix the
  y-axis limits across panels in the plot (#1349).

- Users can now find `plot()` method documentation through `?plot` (#1349). 

- `env_get()` refactored to allow access of R objects inside the model 
  environment similar to `base::get()`; `env_get_obj()` is an alias to `env_get()`, 
  analogous to `env_get_env()` (#1355).

## Bugs Fixed

- Fixed a bug where the model row counter was not correct when a simulation
  was performed with only an idata set (#1351).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# mrgsolve (development version)
+# mrgsolve 2.0.0
 
 - Closed-form three-compartment linear models with or without depot compartments
   can now be implemented via `$PKMODEL` (#1345). 
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,7 +56,7 @@
   - `convert_semicolons()` inserts missing trailing semicolons on assignment
     statements.
   - RStudio addins are provided to allow conversion of certain NONMEM source 
-    blocks code to mrgsolve format in the RStudio editor window.
+    code blocks to mrgsolve format in the RStudio editor window.
 
 - Closed-form three-compartment linear models with or without depot compartments
   can now be implemented via `$PKMODEL` (#1345).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,16 +1,93 @@
 # mrgsolve 2.0.0
 
+## Breaking Changes
+
+- `data_set()` and `idata_set()` no longer accept `.subset`, `.select`,
+  `object`, or `need` arguments. These arguments allowed filtering and
+  column selection inside the call; that processing should now be done on the
+  data frame before passing it to `data_set()` or `idata_set()`.  (#1374)
+
+- The `n` argument to `simeta()` and `simeps()` has been discouraged for a
+  while and is now removed. Calling `simeta(n)` or `simeps(n)` to resimulate 
+  a single ETA or EPS value is no longer supported. Use `simeta()` or `simeps()` 
+  (no argument) to resimulate all values. (#1373)
+
+- `knobs()`, `wf_sweep()`, and `render()` have been removed. These
+  functions were previously deprecated. (#1369)
+
+- The default value of `root` in `$NMXML` and `$NMEXT` has changed from
+  `"working"` to `"cppfile"`. The root directory for locating NONMEM
+  output files now defaults to the directory containing the model `.cpp` file
+  rather than the R working directory. Pass `root = "working"` to restore the
+  previous behavior; but users are _highly_ encouraged to use the new default.
+  (#1368)
+
+- `summarise.each()` is removed; it has not been practically reachable by the 
+  user for several years (#1352).
+
+- Several `modlib` model library models have had parameter and compartment
+  names standardized (#1361):
+  - The first extravascular compartment is now named `EV` (was `EV1`) in
+    `pk1cmt`, `pk2cmt`, `irm1`-`irm4`, and `emax`.
+  - The first absorption rate constant is now named `KA` (was `KA1`) in
+    models that previously used numbered names.
+  - Code that references these compartments or parameters by name (e.g.,
+    `init(mod, EV1 = 0)` or `param(mod, KA1 = 1)`) will need to be updated.
+  - We continue to discourage users from using these models in production 
+    code. 
+
+## New Features
+
 - Closed-form three-compartment linear models with or without depot compartments
-  can now be implemented via `$PKMODEL` (#1345). 
+  can now be implemented via `$PKMODEL` (#1345).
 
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,17 +1,94 @@
...
+  simulation finishes (#1353).
+
+- The `plot()` method for `mrgsims` objects gains a `fixy` argument to fix the
+  y-axis limits across panels in the plot (#1349).
+
+- Users can now find `plot()` method documentation through `?plot` (#1349). 
+
+- `env_get()` refactored to allow access of R objects inside the model 
+  environment similar to `base::get()`; `env_get_obj()` is an alias to `env_get()`, 
+  analagous to `env_get_env()`(#1355).
```
kyleam: typo: analogous

NEWS.md
```diff
@@ -1,17 +1,94 @@
...
+
+## Bugs Fixed
+- Fixed a bug where the model row counter was not correct when a simulation 
+  was preformed with only an idata set (#1351).
```
kyleam: `s/preformed/performed/`, I think

NEWS.md
```diff
@@ -1,17 +1,94 @@
...
+
+- New DSL preprocessing functions convert NONMEM/Fortran model code to C++
+  syntax (#1360, #1362, #1366):
+  - `convert_pow()` converts Fortran `**` exponentiation to `pow()`.
+  - `convert_fort_if()` converts Fortran `IF/THEN/ELSE/END IF` blocks to C++
+    `if/else` syntax.
+  - `convert_semicolons()` inserts missing trailing semicolons on assignment
+    statements.
+  - RStudio addins are provided for each conversion, as well as a combined
+    "Convert NM" addin that applies all three in the correct order.
```
kyleam: Perhaps worth moving all three of these to the top of this section to highlight them as things users may be particularly interested in.

kyleam: > - RStudio addins are provided for each conversion, as well as a combined
>    "Convert NM" addin that applies all three in the correct order.

Two things look off to me there:
 * I'd interpret that as addins for each separate conversion and a combined one, but there are only three: `addin_convert_nm` (IF and semi), `addin_add_semicolons`, and `addin_convert_pow`
 * the "Convert NM" addin doesn't do all three because it doesn't do pow

https://github.com/metrumresearchgroup/mrgsolve/blob/fe4914340f3db6ecfdd2e96d3fc8203b92bb4790/R/dsl-preprocess-addin.R#L28-L29

kylebaron: Moved.

NEWS.md
kylebaron: Thanks; not a good preformance.

NEWS.md
kylebaron: Just gave too much detail there; I'm going to say that the addins let you convert code in place and have the users discover from there.

kyleam (approved):

PR #1383, opened 2026-05-16
Rcpp helper to set names in place

kylebaron:
Some profiling showed that simulated data have been getting copied when I've been setting output data frame names. This is a well known problem but we've been doing it for years now. 

This addition is a simple way to set attributes in place.


kyleam (approved):
Failing CI job is not due to changes in this PR: gh-1384

PR #1387, opened 2026-05-18
defer mat2df until return from DEVTRAN

kylebaron:
I've been thinking a lot about memory management and anything simple we could do to be less careless about copies and the like. 

This PR moves the `mat2df()` call to outside of `DEVTRAN()`. It is functionally no change in behavior. When we do `mat2df()`, there will be a 2x spike in memory utilization because we have both the matrix and the data frame going at the same time. I think this is unavoidable; and I'm convinced saving to matrix won't be beat and we will have to get to data frame to do anything useful with the data. But doing this in `DEVTRAN()`, all of the data records that were created for the simulation are still in scope so that 2x spike is on top of all of that. By moving the `mat2df()` call outside of `DEVTRAN()`, this will let all the data record objects get torn down prior to the 2x spike. 

QUESTION: should I update `mat2df()` to pass in data frame names so we can set them in the same call? or just leave it as is for now (with calls to both `mat2df()` and `set_names_inplace()`?

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -95,12 +95,8 @@ EXPAND_OBSERVATIONS <- function(data, times, to_copy, next_pos) {
     .Call(`_mrgsolve_EXPAND_OBSERVATIONS`, data, times, to_copy, next_pos)
 }
 
-set_names_inplace <- function(x, nm) {
-    invisible(.Call(`_mrgsolve_set_names_inplace`, x, nm))
-}
-
-mat2df <- function(x) {
-    .Call(`_mrgsolve_mat2df`, x)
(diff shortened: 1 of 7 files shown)


kyleam:
> QUESTION: should I update `mat2df()` to pass in data frame names so we can set them in the same call? or just leave it as is for now (with calls to both `mat2df()` and `set_names_inplace()`?

I like the idea of doing it in the same call given the current one-to-one coupling (each `mat2df` calls is followed by a call to `set_names_inplace` and each `set_names_inplace` call is preceded by `mat2df`).  But keeping it separate is also fine by me.

kyleam (approved):

Issue #1396, opened 2026-06-09
EVID==4 doesn't reset when there's a lagtime


kylebaron:
Pretty sure I've seen this before. 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '
$PARAM @annotated @input
CL   :  1 : Clearance (volume/time)
V    : 20 : Central volume (volume)
KA   :  1 : Absorption rate constant (1/time)
LAG  :  0 : Lagtime

$CMT  @annotated
EV   : Extravascular compartment
CENT : Central compartment

$GLOBAL
#define CP (CENT/V)

$PKMODEL advan = 2

$MAIN 
ALAG_EV = LAG;

$CAPTURE @annotated
CP : Plasma concentration (mass/volume)
'

mod <- mcode("code", code) %>% param(KA = 0.05, CL = 0.8)
#> Building code ...
#> done.

e <- ev(amt = 100, ii = 24, addl = 7) %>% realize_addl() %>% as_data_set()
e$evid[4] <- 4

mod <- param(mod, LAG = 0)

out <- mrgsim(mod, data = e, delta = 0.01, end = 240)

plot(out)
```

![](https://i.imgur.com/hoso9Z0.png)

``` r


mod <- param(mod, LAG = 0.1)

out <- mrgsim(mod, data = e, delta = 0.01, end = 240)

plot(out)
```

![](https://i.imgur.com/n6miQxb.png)

<sup>Created on 2026-06-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

kylebaron:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(mrgsolve)
#> 
#> Attaching package: 'mrgsolve'
#> The following object is masked from 'package:stats':
#> 
#>     filter

code <- '
$PARAM @annotated @input
CL   :  1 : Clearance (volume/time)
V    : 20 : Central volume (volume)
[...]

PR #1397, opened 2026-06-10
Fix bug where system didn't reset on EVID 4 with a lag time

kylebaron:
There was a logic issue when processing EVID==4 records when the record was already "unarmed".  The flawed logic was introduced when trying to match what NONMEM does on EVID=4 with a lag time.  #1262

I saw a bug in production code where the system was not correctly reset prior to reset/dose when there was a lag time. The logic in #1262 did not properly deal with the situation where the dose had a lag time.

This PR adds a flag to `odeproblem` called `system_at_ss`; whenever we advance the system to steady state, we set this flag to `true`. Then, when processing an `EVID=4` record, we decline to reset the system of we find we're already at ss.  With this check in place, the EVID=4 logic works. 

See #1396 for before / after fix.

### DIFF
--- a/inst/maintenance/unit/test-z-alag-f.R
+++ b/inst/maintenance/unit/test-z-alag-f.R
@@ -48,11 +48,29 @@ mod <-
   carry_out(evid) %>% 
   update(end=2)
 
-test_that("ss=1 and F_CMT =0 issue-497", {
+test_that("ss=1 and F_CMT=0 issue-497", {
   mod <- param(mod, F1 = 0)
   out <- mrgsim(mod, ev(amt = 100, ii = 24, ss=1), delta=1,end=6)
   expect_is(out, "mrgsims")
   expect_true(all(out$CENT==0))
+})
+
+test_that("EVID=4 with lag time still resets", {
+  mod <- param(mod, ALAG1 = 0.25)
+  e <- ev_seq(
+    ev(amt = 100, ii = 12, addl = 3), 
+    wait = 24, 
+    ev(amt = 50, evid = 4), 
+    wait = 24,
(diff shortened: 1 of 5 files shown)


src/datarecord.cpp
```diff
@@ -191,16 +191,18 @@ void datarecord::implement(odeproblem* prob) {
...
+    if(this->ss()==0 && !prob->system_at_ss) {
       for(int i=0; i < prob->neq(); ++i) {
         prob->y(i,0.0);
         prob->on(i);
         prob->rate0(i,0.0);
       }
       prob->init_call(Time);
     }
+    // For example, doses at lag times
+    if(!Armed) break;
```
kylebaron, line 205: This move lets a lagged dose reset the system at the time of the parent dose, but no reset if we already brought the system to steady state with `ss=1`.

inst/maintenance/unit/test-z-alag-f.R
```diff
@@ -48,13 +48,32 @@ mod <-
...
 
-test_that("ss=1 and F_CMT =0 issue-497", {
+test_that("ss=1 and F_CMT=0 issue-497", {
   mod <- param(mod, F1 = 0)
   out <- mrgsim(mod, ev(amt = 100, ii = 24, ss=1), delta=1,end=6)
   expect_is(out, "mrgsims")
   expect_true(all(out$CENT==0))
 })
 
+test_that("EVID=4 with lag time still resets gh-1396", {
```
kyleam, line 58: Looks good, and confirmed that this test fails if code changes are reverted.

kyleam (approved):

Issue #1395, opened 2026-06-08
nocb unexpected behavior (I think)


curtisKJ:
Hello,

I've noticed that when simulating with a dose-dependent relative F, the advancing behavior for nocb is not what I would have assumed it to be. In the example below, three doses are administered, the first two doses are of 1000 mg with a lower relative F, and the final third dose is of 100 mg, having the reference relative F of 1. I did a quick search and didn't find anything exactly like this, but apologies if I missed it. 

My expectation was the following:

- Without nocb, the second dose would have an incorrect relative F applied to PK records after the second dose due to DOSE coming backwards from the next (third) dose record. Using nocb would resolve this, I thought...

- However, employing nocb seems to cause the preceding, lower relative F from the 1000 mg dose to be employed to the third dose of 100 mg, despite the DOSE being correctly listed in the output dataset for the EVID =1 record at that time (see output at end of code snippet). Adjusting recsort does not impact the behavior. 

- I would have thought that having explicit dosing records with DOSE listed would supersede any nocb comings and goings. Are my assumptions flawed.....Or is it perhaps the output I've generated is not accurately showing what was used when the dose was realized in the system? .........Or - is this because I'm using a decrepit version of mrgsolve (mrgsolve_1.1.1) from an _**old**_ repo?

Thanks,

-Curtis

```
library(mrgsolve)
library(dplyr)
library(tidyr)
library(ggplot2)

# ------------------------------------------------------------------------------
# 1-CMT oral model using $PKMODEL 
# Dose-dependent event bioavailability:
#   - dose 1000 -> F_GUT = 0.5
#   - dose 100  -> F_GUT = 1.0
#
# Day 1:  1000 mg
# Day 8:  1000 mg
# Day 15:  100 mg
#
# DOSE is only populated on dose rows so you can compare nocb = TRUE vs FALSE.
# ------------------------------------------------------------------------------

code <- '
$PARAM CL = 5, V = 50, KA = 0.015, DOSE = 100

$CMT GUT CENT
$PKMODEL ncmt = 1, depot = TRUE

$MAIN
double DOSEF = 1.0;
if(DOSE == 1000) DOSEF = 0.5;

F_GUT = DOSEF;

$TABLE
capture CP = CENT / V;
capture DOSE    = DOSE;
capture DOSEFN   = DOSEF;
capture GUT_AMT = GUT;
capture CENT_AMT = CENT;
'

[...]

PR #1399, opened 2026-06-21
Adjust parameter update for doses under locf

kylebaron:
There was a user report where bioavailability was derived from a data set item wasn't behaving as expected. Because bioav comes from the data, it means that the results will be sensitive to `nocb` / `locf` behavior. 

https://github.com/metrumresearchgroup/mrgsolve/issues/1395

I did a simulation against nonmem and found that the `nocb` case was correctly handled. The principle was that the value of the parameter on the data set row for the dose record is what determines F in this case. 

There is a demonstration here: https://github.com/mrgsolve/nmtests/blob/master/nocb/1001.md

It's not obvious if it should be different or the same when you simulate with `locf`.  As a simple solution, I implemented a change where we follow the same behavior for `locf` (it's the value of the parameter on the dosing record that determines F). To do this, we have to call `$MAIN` again, after the `locf`-targeted parameter update. 

This PR also changes the name of `copy_next_parameters()` to `copy_parameters_nocb()` to better reflect the fact that this handles the `nocb` case only.

### DIFF
--- a/inst/maintenance/unit/test-nocb.R
+++ b/inst/maintenance/unit/test-nocb.R
@@ -151,3 +151,65 @@ $CAPTURE DV, CL, COV2 = COV
 })
 
 
+nocb_doses <- '
+$PARAM CL = 1, V = 20, KA = 1.2, FLAG = 0
+$PKMODEL advan = 2
+$PK
+F_A1 = 1; 
+if(FLAG==1) F_A1 = 0.5;
+capture FF = F_A1;
+$CAPTURE FLAG
+'
+mod <- mcode("nocb-doses", nocb_doses)
+ 
+test_that("Covariate effects on bioavailability come on the dose record", {
+ 
+  data <- data.frame(
+    ID = 1,
+    TIME = c(0,24,48,72), 
+    EVID = 1, 
+    AMT = 1000, 
+    FLAG = c(0, 1, 1, 0), 
+    CMT = 1
+  )
+  
+  mod <- update(mod, end = 168)
+  
+  out1 <- mrgsim(mod, data, carry_out = "evid")
+  out2 <- mrgsim(mod, data, carry_out = "evid", nocb = FALSE)  
+  out1 <- filter_sims(out1, evid==1)
+  out2 <- filter_sims(out2, evid==1)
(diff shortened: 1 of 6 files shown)


src/devtran.cpp
```diff
@@ -524,6 +508,12 @@ Rcpp::List DEVTRAN(const Rcpp::List parin,
         prob.eps(k,eps(crow,k));
       }
       
+      // nocb we copy tto parameters just prior to $PK
+      if(dat.any_copy && (nocb)) {
```
kyleam: Unnecessary parens around `nocb`?

inst/maintenance/unit/test-nocb.R
```diff
@@ -151,3 +151,77 @@ $CAPTURE DV, CL, COV2 = COV
...
+    CMT = 1
+  )
+  
+  mod <- update(mod, end = 168)
+  out1 <- mrgsim(mod, data, carry_out = "evid,DFLAG = FLAG")
+  out2 <- mrgsim(mod, data, carry_out = "evid,DFLAG = FLAG", nocb = FALSE) 
+  
+  ## Covariate nocb
+  expect_equal(out1$FF[c(1,2,3,4)], c(1, 1, 0.5, 0.5))
+  expect_identical(out1$FLAG, out1$FLAG)
```
kyleam: `out1$FLAG` is repeated.  Should the second one be `out1$DFLAG`?

kyleam (approved):
Thanks for the background.  Seems reasonable to me as far as I can understand.

inst/maintenance/unit/test-nocb.R
kylebaron: Yes; thanks.

kylebaron:
@kyleam I added some more tests that may not fail if this change is reverted, but are in there to clarify / confirm nocb / locf behavior.

kyleam:
The 4.1 and 4.2 jobs have started to fail.  Should be unrelated to any of the changes here.  I'll look into those.

kyleam:
> The 4.1 and 4.2 jobs have started to fail. Should be unrelated to any of the changes here. I'll look into those.

CI is all set.

